PR #234, opened 2026-05-27
ci: fix release dispatch — actions:write, docker tag push, makemkv marker

Jsakkos:
## Summary

Three follow-up fixes for the release CI chain, discovered during the v0.8.0 release run.

- **`tag-release.yml`** — adds `actions: write` permission so `GITHUB_TOKEN` can dispatch other workflows via `gh workflow run` (was 403-ing without it); also dispatches `docker.yml` alongside `release.yml` so the container image is published on every tag push
- **`docker.yml`** — adds a `workflow_dispatch` trigger and `IS_RELEASE` environment flag so the image can be built and pushed to GHCR when dispatched manually (e.g. for backfill) or from `tag-release`; passes the explicit tag through `checkout ref` and `docker/metadata-action` tag patterns so the correct semver label is applied
- **`makemkv-install.yml`** — switches the post-install version check from running `makemkvcon --version` (no such flag; exits non-zero) to reading the `.installed-version` marker file written by `install-makemkv.sh`, fixing spurious "MakeMKV install check" CI failures

## Root cause

`GITHUB_TOKEN` in `tag-release.yml` lacked the `actions: write` permission scope, which GitHub requires for `workflow_dispatch` events. The `actions: write` grant is the minimal fix; an alternative (`workflow_call` restructure) was explored in PR #225 but landed after that PR had already merged.

## Test plan
- [x] Pre-commit hooks pass (yaml, trim whitespace)
- [x] `check-yaml` passes on all three modified workflow files

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,6 +3,12 @@ name: Docker
 on:
   push:
     tags: ["v*"]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Git tag to build and push to GHCR (e.g. v0.8.1). Leave empty for a smoke-test-only branch build.'
+        required: false
+        default: ''
   pull_request:
     paths:
       - "Dockerfile"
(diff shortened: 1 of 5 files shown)


Jsakkos:
Superseded by #230, #231, and #232, which landed these same three fixes independently. No unique content remains — closing.

PR #289, opened 2026-06-02
fix: retry job creation for disc stranded during active job

katelovescode:
## Problem

When a new disc is inserted while a job for that drive is in a blocking state (IDENTIFYING, RIPPING, MATCHING, ORGANIZING), `_create_job_for_disc` finds the active job and returns early without creating a new one. The sentinel updates `_drive_states[drive]` to `True` for the new disc, but once the blocking job completes, no retry fires — the sentinel won't re-emit an "inserted" event because the disc state hasn't changed from its perspective. The new disc is permanently stranded until the drive is physically removed and reinserted, or until the process restarts.

## Fix

- **`_retry_disc_if_present(job_id)`** — loads the job's `drive_id`, calls `is_disc_present` (returns `False` safely for non-drive IDs like `"staging"`), and delegates to `_create_job_for_disc` if a disc is found. All existing guards (15s cooldown, duplicate job check) apply unchanged.
- **`_on_job_terminal`** — registered with `state_machine.on_terminal_state` to handle COMPLETED and FAILED transitions.
- **`_on_state_transition_retry`** — registered with `state_machine.on_transition` to handle REVIEW_NEEDED. REVIEW_NEEDED unblocks the drive (it's excluded from `_create_job_for_disc`'s blocking query) but doesn't fire terminal callbacks. Uses `self._loop.create_task` to bridge the sync callback contract, following the existing pattern in this codebase.

All changes are contained in `backend/app/services/job_manager.py`. No changes to `job_state_machine.py` or `sentinel.py`.

## Tests

Three unit tests in `tests/unit/test_job_manager.py` (`TestRetryDiscIfPresent`):
- disc present → `_create_job_for_disc` called with correct drive and label
- disc absent → no-op
- unknown job ID → no-op

### DIFF
--- a/backend/app/core/sentinel.py
+++ b/backend/app/core/sentinel.py
@@ -292,6 +292,17 @@ class DriveMonitor:
 
         logger.info("Drive monitor stopped")
 
+    def notify_ejected(self, drive: str) -> None:
+        """Reset drive state after a programmatic eject.
+
+        eject_disc() only issues the OS command; it doesn't update _drive_states.
+        Without this reset, if a new disc is inserted before the poll detects the
+        removal, the sentinel sees True → True and fires no 'inserted' event,
+        leaving the new disc permanently stranded.
+        """
+        self._drive_states[drive] = False
+        self._pending_changes.pop(drive, None)
+
     async def _poll_loop(self) -> None:
         """Poll for drive changes."""
         # Load poll interval from config
--- a/backend/app/services/job_manager.py
+++ b/backend/app/services/job_manager.py
@@ -1533,11 +1533,12 @@ class JobManager:
                                 )
                     await stall_session.commit()
 
-            # Eject disc
+            # Eject disc and reset sentinel state so a new disc insert is detected
             try:
                 from app.core.sentinel import eject_disc
 
                 await asyncio.to_thread(eject_disc, drive_id)
+                self._drive_monitor.notify_ejected(drive_id)
             except (OSError, RuntimeError) as e:
                 logger.warning(f"Could not eject disc from {drive_id}: {e}")
 
--- a/backend/tests/unit/test_job_manager.py
+++ b/backend/tests/unit/test_job_manager.py
@@ -422,3 +422,44 @@ class TestOnePassRipFallback:
         assert rip.await_count == 2
         assert rip.await_args_list[0].kwargs["title_indices"] is None  # one-pass
         assert sorted(rip.await_args_list[1].kwargs["title_indices"]) == [0, 1]  # fallback
+
+
+@pytest.mark.unit
+class TestRunRippingCallsNotifyEjected:
+    async def test_notify_ejected_called_after_rip(self, rip_env, monkeypatch):
+        """_run_ripping must call notify_ejected on the drive monitor after ejecting."""
+        from unittest.mock import MagicMock
+
+        job, _ = await _seed(content_type=ContentType.TV, staging=str(rip_env), is_selected=True)
+        monkeypatch.setattr(job_manager, "_backfill_unmatched_titles", AsyncMock())
+        _mock_rip(monkeypatch, RipResult(success=True, output_files=[]))
+
+        spy = MagicMock()
+        monkeypatch.setattr(job_manager._drive_monitor, "notify_ejected", spy)
+
+        await job_manager._run_ripping(job.id)
+
+        spy.assert_called_once_with(job.drive_id)
+
+
+@pytest.mark.unit
+class TestNotifyEjected:
+    def test_resets_drive_state_and_clears_pending(self):
+        """notify_ejected resets _drive_states and clears any pending debounce."""
+        from app.core.sentinel import DriveMonitor
+
+        monitor = DriveMonitor()
+        monitor._drive_states["/dev/sr0"] = True
+        monitor._pending_changes["/dev/sr0"] = 1
+
+        monitor.notify_ejected("/dev/sr0")
+
+        assert monitor._drive_states["/dev/sr0"] is False
+        assert "/dev/sr0" not in monitor._pending_changes
+
+    def test_no_error_when_drive_not_tracked(self):
+        """notify_ejected on an untracked drive is a no-op, does not raise."""
+        from app.core.sentinel import DriveMonitor
+
+        monitor = DriveMonitor()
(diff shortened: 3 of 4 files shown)


katelovescode:
Side note, last night I had a disc rip all the way to the end and I resolved 6 duplicate tracks and hit "process" - they finished successfully.  This morning I put in another disc and it didn't pick it up automatically.  I'm running in debug mode so I just fired `curl -s -X POST http://localhost:8010/api/simulate/trigger-real-scan | python3 -m json.tool`, but I think this PR will resolve that case, too.  Sentinel can *see* the drive, but it doesn't pick it up, even after a full "complete" cycle.

```
curl -s http://localhost:8010/api/drives | python3 -m json.tool
[
    {
        "drive_id": "/dev/sr0",
        "status": "ready"
    }
]
```

Jsakkos:
@claude review this PR

katelovescode:
"edge case" Claude, ouch 😂 Sequential disc burning!  Everybody wants that! 😉

Jsakkos:
> Side note, last night I had a disc rip all the way to the end and I resolved 6 duplicate tracks and hit "process" - they finished successfully. This morning I put in another disc and it didn't pick it up automatically. I'm running in debug mode so I just fired `curl -s -X POST http://localhost:8010/api/simulate/trigger-real-scan | python3 -m json.tool`, but I think this PR will resolve that case, too. Sentinel can _see_ the drive, but it doesn't pick it up, even after a full "complete" cycle.
> 
> ```
> curl -s http://localhost:8010/api/drives | python3 -m json.tool
> [
>     {
>         "drive_id": "/dev/sr0",
>         "status": "ready"
>     }
> ]
> ```

I've seen this too, intermittently. Thanks for the fix.

katelovescode:
Do you want me to rebase this to auto-merge or do you have automations to do that?

Jsakkos:
Thanks for digging into this, @katelovescode — the stranded-disc behavior is a real and annoying bug, and the root-cause write-up in the description is accurate: the sentinel is edge-triggered and won't re-emit `inserted` for a disc it already tracks as present.

I did a deep pass on the implementation and I think the retry needs to move down a layer before we merge. The core issue is that the sentinel is **edge-triggered** (`_drive_states` in `sentinel.py`), but this fix adds a **level check** ("is a disc physically present right now?") at every job-state transition. Those two models disagree about "the same disc still sitting in the tray," and the only thing currently reconciling them is the post-rip eject — so any path that *doesn't* eject re-opens the gap. That leads to a few concrete problems:

**1. Duplicate-job loop on `REVIEW_NEEDED`.** An ambiguous disc goes to `REVIEW_NEEDED` during *identification, before any rip* (`identification_coordinator.py:301/365/441`), so it's never ejected. `_create_job_for_disc`'s existing-job guard *excludes* `REVIEW_NEEDED` (`job_manager.py:348-350`), so once the 15s cooldown lapses (`:360`) the retry creates a fresh job for the same disc → re-identifies → `REVIEW_NEEDED` again → repeat. There's no `content_hash`/label "already-processed" check, so only physically ejecting the disc breaks the cycle.

**2. Two `makemkvcon` on one drive.** While a job waits in `REVIEW_NEEDED`, the retry can spawn a phantom `IDENTIFYING` job that scans the same drive (`identification_coordinator.py:144`); when the reviewer hits approve, the original job starts ripping the same physical drive (`routes.py:815-821` → `job_manager.py:1438`). The per-drive lock only spans the DB insert, not the scan/rip, so the two contend for one optical drive — the conflict CLAUDE.md specifically warns about.

**3. Early-`FAILED` loop.** A disc that fails identification ("No titles found", MakeMKV errors — `identification_coordinator.py:148/156/478`) also isn't ejected, and `FAILED` is excluded from the guard too, so `_on_job_terminal` re-creates a job that fails the same way ~every 15s until the disc is pulled.

**4. It may not fix the reported case.** The repro in the description — a *new* disc inserted into an idle drive after a completed cycle — doesn't involve any job transition, so none of the new callbacks fire on it. The disc that's actually stranded is the one inserted while nothing is happening.

I think the durable fix is in `sentinel.py` rather than at the job layer:

- Reset `_drive_states[drive]` on eject and on confirmed removal (right now `eject_disc` never touches it), and/or add a periodic idle reconcile so a missed removal edge self-heals. That directly addresses the "idle drive doesn't pick up the next disc" symptom.
- If you want to keep a completion-time retry as a belt-and-suspenders, it'd need to (a) fire on `COMPLETED` only, (b) skip when the present disc is the same one the job just processed (compare `content_hash`/label, or require an absent→present edge since), and (c) offload `is_disc_present`/`get_volume_label` via `asyncio.to_thread` — they're blocking ctypes/`blkid` calls and currently run inline on the event loop inside the awaited transition.

One more on the tests: `TestRetryDiscIfPresent` mocks out `_create_job_for_disc` and calls `_retry_disc_if_present` directly, so it passes without exercising the dedup/cooldown/transition wiring — a test that drives a real `REVIEW_NEEDED`/`COMPLETED` transition (without mocking `_create_job_for_disc`) would catch the loop above.

Happy to pair on the sentinel side if useful — the diagnosis here is good, it's mostly a question of which layer the reconcile lives in.

Jsakkos:
> Do you want me to rebase this to auto-merge or do you have automations to do that?

You'll need to rebase after reviewing the comments above and resolving or discussing.

katelovescode:
All four points verified against the code. Reworked.

**What changed:**

The job-layer callbacks are gone. The fix is now entirely in `sentinel.py`: added `DriveMonitor.notify_ejected(drive)` which resets `_drive_states[drive]` to `False` and clears any pending debounce state. `job_manager.py` calls it immediately after the post-rip `eject_disc()` call.

This directly addresses the root cause: `eject_disc()` issued the OS command but never updated `_drive_states`, so a disc inserted before the next poll cycle was seen as `True → True` with no edge to fire `inserted`. With this fix, the sentinel always sees `False → True` on the next poll and fires correctly.

The REVIEW_NEEDED loop, two-makemkvcon, FAILED loop, and blocking-calls-on-event-loop issues all go away with the callbacks removed. The original reported case (new disc stranded in an idle drive after a completed cycle) is handled by the sentinel reset.

Tests updated: removed `TestRetryDiscIfPresent`, added `TestNotifyEjected` covering reset behavior and untracked-drive no-op.

PR #294, opened 2026-06-02
feat(dashboard): surface TMDB health warning as banner + per-job inline alert

### LINKED ISSUE #262: UX: TMDB onboarding follow-ups — dashboard health banner (P2) + job-level error surfacing (P3)
Follow-up to #243.

The config-wizard redesign addresses the wizard-scoped parts of #243:

- [x] **P0** — instructional placeholder (the field no longer shows a token-shaped `eyJhbGci…` default that reads as "already filled")
- [x] **P1** — auto-validate the token on blur (inline ✓/✗, no button hunt) + first-run gate (onboarding can't pass the TMDB step without a validated token or an explicit "continue without token")

The remaining parts of #243 touch other layers and are tracked here:

## P2 — Startup / persistent health check (frontend)

On app load, if `tmdb_api_key` is not configured, show a **dismissible** warning banner on the dashboard, reusing the existing banner pattern in `frontend/src/app/App.tsx` (update banner / platform banner). Optionally re-validate the stored token once per session and warn if it has expired or been revoked.

## P3 — Error surfacing during jobs (cross-stack)

When TMDB classification fails because the key is absent or invalid, set a human-readable field on the job (or broadcast via WebSocket) so the DiscCard / error panel can show:

> _"TMDB API key not configured — classification running in heuristic-only mode. Configure your token in Settings."_

Likely files: `backend/app/core/tmdb_classifier.py`, `backend/app/services/identification_coordinator.py`, the `DiscJob` model, and `DiscCard`.

## Acceptance

- [ ] Dashboard shows a persistent (dismissible) banner if TMDB is not configured at startup
- [ ] Job detail / error panel names TMDB misconfiguration explicitly when it's the cause of degraded classification

katelovescode:
Fixes #262

## Summary

- Adds a dismissible amber banner below the nav when no TMDB API key is configured, with an animated enter/exit via Framer Motion and a direct link to open the settings wizard
- Adds an inline warning inside each active `DiscCard` when TMDB is unconfigured, using the existing `isActive` guard so completed/failed jobs stay uncluttered
- Both warnings clear automatically after a token is saved — the config is re-fetched on settings wizard close without requiring a page reload

## Test Plan

- [ ] Load dashboard with no TMDB token set — amber banner visible below nav
- [ ] Click "Configure token" in banner — settings wizard opens
- [ ] Click X on banner — banner animates out, does not return until page reload
- [ ] Simulate an active job — DiscCard shows inline amber warning; completed jobs do not
- [ ] Save a TMDB token — both banner and card warning disappear without a page reload


Jsakkos:
@claude review this PR

katelovescode:
Pushed a fix for the review comments

Jsakkos:
🎉 Thank you for this, @katelovescode — your first contribution to Engram shipped in v0.15.0 (the TMDB health-warning banner). We've added you to CONTRIBUTORS.md and credited you in the release notes. Welcome aboard! 🚀

PR #326, opened 2026-06-04
chore(ci): fix Dependabot auto-merge + ignore Vite major

Jsakkos:
## Why

Ten Dependabot PRs piled up open. Root cause: the `dependabot-auto-merge.yml` workflow's **approve step crashes** with `GitHub Actions is not permitted to approve pull requests`, which aborts the job *before* `gh pr merge --auto` runs — so minor/patch bumps never auto-merged. Branch protection on `main` requires **no approving review**, so the step is both unnecessary and harmful.

## What

- **Remove the broken `gh pr review --approve` step**; keep `gh pr merge --auto --squash`. Minor/patch Dependabot PRs will auto-merge again.
- **Bump `dependabot/fetch-metadata` 2 → 3** in the same file — **supersedes #218** (which only made this one-line change and conflicts with this PR).
- **Ignore Vite *major* bumps** in `dependabot.yml`. Vite 8 is a coordinated, dev-only upgrade (`vite` + `@vitejs/plugin-react@6` + `@tailwindcss/vite@4.3.0` must move together); Dependabot raises a vite-only PR that fails `npm ci` with `ERESOLVE` (that's #220). The upgrade should be done deliberately as its own PR.

## Follow-up (handled separately, not in this PR)

- Merge the 4 green GitHub-Actions PRs (#312, #313, #314, #315).
- Rebase + merge the 4 safe npm PRs (#219, #221, #222, #223).
- Close #218 (superseded here) and #220 (deferred via the ignore above).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,13 @@ updates:
     directory: "/frontend"
     schedule:
       interval: "weekly"
+    ignore:
+      # Vite major bumps are a coordinated, dev-only upgrade (vite +
+      # @vitejs/plugin-react + @tailwindcss/vite must move together) that
+      # Dependabot can't do atomically — it raises a vite-only PR that fails
+      # `npm ci` with ERESOLVE. Keep majors off; do the upgrade deliberately.
+      - dependency-name: "vite"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: "pip"
     directory: "/backend"
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -13,19 +13,14 @@ jobs:
     steps:
       - name: Fetch Dependabot metadata
         id: metadata
-        uses: dependabot/fetch-metadata@v2
+        uses: dependabot/fetch-metadata@v3
         with:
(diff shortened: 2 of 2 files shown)


Jsakkos:
Thanks for the review. Both suggestions are non-blocking; I've considered each and am intentionally not changing them:

**Parallel ignores for `@vitejs/plugin-react` / `@tailwindcss/vite` majors** — keeping the ignore scoped to `vite` on purpose. A failing `@vitejs/plugin-react` v5/v6 PR is the signal we *want*: it's what should prompt the deliberate, coordinated Vite-8 upgrade (vite + plugin-react + @tailwindcss/vite together). Pre-emptively ignoring those majors would hide that prompt and suppress updates that may resolve cleanly. As you noted, this is best handled reactively — I'll add a parallel entry only if/when one of those actually 502s on ERESOLVE.

**SHA-pin `fetch-metadata@v3`** — every GitHub Action in this repo uses floating major tags (it's exactly what the open Dependabot github-actions PRs bump). Pinning one action to a full SHA would break that convention and Dependabot's tag-based updates for it. For a trusted first-party action I'd rather keep it consistent with the rest of the workflows.

No code change from this review.

PR #339, opened 2026-06-05
fix(import): resolve tmdb_id for generic-label imports so matching finds references

Jsakkos:
## Problem

Watch-folder imports (`drive_id="import"`) of a TV show were dumping **every** track into **Needs Review** with no real match, the review panel showed **no episode list**, and bonus features weren't tagged as Extras. Reproduced on a 7-season **Seinfeld** import (all jobs `tmdb_id=null`, every track in review).

## Root cause

An import derives `detected_title`/`detected_season` from the folder tree (`…\Seinfeld\Season 3\`), but the identify-time TMDB lookup in `_run_classification` is gated on the **volume label** — which for an import is the generic `"SEASON_3"` → no name → lookup skipped → **`job.tmdb_id` is persisted as null**.

Since #288 the subtitle/reference cache is **tmdb-id-keyed** on disk (`~/.engram/cache/data/<tmdb_id>/`). The matcher builds its reference dir via `corpus_dir_name(expected_tmdb_id, show_name)`, which falls back to the **name** when the id is null → `data/Seinfeld` (no longer exists post-#288) → `identify_episode` logs `No reference subtitle files found` → ASR never runs. Every track then degrades to the curator's filename guess (`confidence=0.3` if the filename has an episode code, else `0.0`), all below the 0.7 auto-organize gate → review. The same null id makes `GET /api/jobs/{id}/season-roster` return `available:false` (blank episode list) and weakens the duration-based extras pre-filter.

It's a regression: pre-#288 the cache was name-keyed, so null-id imports still matched by name.

## Fix

New `IdentificationCoordinator._resolve_missing_tmdb_id` resolves + persists `tmdb_id` from the job's known title via the existing `classify_from_tmdb`, wired into:

- **`identify_from_staging`** — right before the same-name-collision block, so the resolved signal also feeds the existing review-routing.
- **`set_name_and_resume`** — the identical bug for generic-label discs named via the NamePromptModal.

Confident single matches auto-resolve and proceed. **Ambiguous / no-year same-name twins** (e.g. *Frasier* 1993 vs the 2023 revival) are left null and routed to review via the existing `ambiguous_identity` / `should_flag_no_year` machinery — the #287 collision safety is preserved. The id is resolved upstream; the matcher already keys on `expected_tmdb_id`.

## Verification

- New unit tests (`tests/unit/test_import_tmdb_resolution.py`), written test-first: confident-resolve, ambiguous→review, and the disc-naming path.
- Full unit suite: **1737 passed**; `ruff check` / `ruff format` clean.
- Read-only check against a real cache: with `tmdb_id=1400`, `corpus_dir_name → data/1400` resolves the season SRTs; with `None`, `data/Seinfeld` is empty (the bug).

## Recovering already-stuck jobs

No re-ripping needed — files are already on disk. Either re-import the folder (runs through the fixed path) or use the existing **Re-identify** action, which already resolves the id from the corrected name.

## Notes for reviewers

- The changelog `(#NNN)` placeholder will be updated to this PR's number.
- A pre-match resolver "chokepoint" in `matching_coordinator` was considered as defense-in-depth but deferred — the import-path fix covers the reported case.


Jsakkos:
Superseded by a fresh PR from the same branch so the **Code Review** workflow runs. It triggers only on `pull_request: opened`, and #339 was opened in a conflicted state (CHANGELOG collision with #338) — GitHub runs no `pull_request` workflows when the open-time merge is unmergeable. Conflict is since rebased/resolved; same commits.

PR #348, opened 2026-06-06
fix(ordering): narrow episode-ordering selector to aired + DVD only

Jsakkos:
## What

The review-queue per-show **episode-ordering selector** could show more than two buttons — one per TMDB episode-group type a show happens to have (digital, story-arc, production, TV) — even though the global **Episode Ordering** setting in the Config wizard only ever offered **Aired** and **DVD**. The two surfaces disagreed; a show with several diverging groups showed up to 6 options in review.

This narrows the selectable set so both surfaces agree: **Aired + DVD only.**

## Why

`ALLOWED_ORDERINGS` in `backend/app/core/episode_ordering.py` is the single gate consumed by:
- `build_ordering_options()` — the review-panel selector (frontend just renders what the backend sends)
- config + per-show API validation (`PUT /api/config`, `PUT /api/shows/{id}/ordering`)
- the per-show ordering resolver

It contained six orderings while the Config UI was deliberately limited to `aired`/`dvd` (#259). Narrowing the frozenset to `{aired, dvd}` brings every surface into lock-step in one change.

The deferred group-type constants and the TMDB `type`→ordering map are **kept** (recognized-but-deferred, mirroring how `absolute` was already handled), so re-enabling any later is a one-token change (add the constant back to the frozenset + the Config UI).

## Reviewer notes

- **Reverses a previously-deliberate decision.** PR #259 intentionally kept the per-show selector wide while the global dropdown stayed narrow. The maintainer confirmed they want both surfaces at aired+DVD, so this reverses that.
- **Invariant untouched:** canonical TMDB **aired** numbering remains the sole identity for matching, job history, and the fingerprint network. This only affects which output orderings a user can *select*.
- Config/per-show API now returns **422** for `digital`/`story_arc`/`production`/`tv` (same as `absolute` already did).

## Tests

- New `TestAllowedOrderingsScope` (allow-list is exactly `{aired, dvd}`; deferred types excluded).
- New `test_only_aired_and_dvd_surfaced_even_when_other_groups_exist` — a show with diverging digital/story-arc/production/tv groups still surfaces only Aired + DVD.
- Extended config API rejection coverage to the deferred orderings.
- Followed TDD (watched all 10 new assertions fail first). Full ordering suite: **58 passed**, ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to Engram will be documented in this file.
 
 ### Fixed
 
+- **The review panel's episode-ordering picker now offers only Aired and DVD order** — the per-show ordering selector in the review queue could show extra buttons (digital, story-arc, production, TV) whenever a show happened to have those alternative orderings on TMDB, even though the global Episode Ordering setting only ever offered Aired and DVD. The two surfaces now agree: both offer just Aired and DVD. The other orderings are still recognized internally but no longer selectable, and matching, history, and the fingerprint network remain on canonical aired numbering as before. (#348)
 - **The "Try LLM match" button in the review Inspector now works** — clicking it on a track in review did nothing useful: the backend failed instantly on a wrong internal import (it looked for an `episode_curator` that doesn't exist), caught the error, and returned a silent success, so the button appeared to do nothing and the failure only showed up in the logs. AI-assisted single-track matching from the review panel now runs the transcription and returns a suggestion as intended. (#347)
 
 ## [0.16.2] - 2026-06-05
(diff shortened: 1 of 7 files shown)


Jsakkos:
Both comment-only items addressed in 82a6896:

- `routes.py:618` — `OrderingOption.ordering` inline comment updated from `# "aired" | "dvd" | "digital" | ...` to `# "aired" | "dvd" (v1 scope)`, matching the equivalent fixes already made in `show_ordering.py` and `types.ts`. This was the one surface left behind.
- `test_episode_ordering.py:72` — `test_ordering_with_no_group_falls_back` comment corrected: the identity result for `digital` now comes from the `ALLOWED_ORDERINGS` short-circuit in `get_projection_map`, not from group absence. Comment now describes the actual mechanism.

No behavior change. The genuine no-group fallback path (allowed ordering, no matching group) remains covered by `test_absent_type_resolves_to_none` at the resolver and `test_no_groups_offers_only_aired`. Ordering suite still green (39 passed), ruff clean.

PR #379, opened 2026-06-10
fix(identify): offload blocking TMDB lookups off the event loop

Jsakkos:
## Summary

Follow-up to [#377](https://github.com/Jsakkos/engram/pull/377), which fixed the same class of bug for `re_identify` and explicitly scoped this path out as a dedicated follow-up.

`IdentificationCoordinator._run_classification` defined a nested **synchronous** helper `_try_tmdb` that called `classify_from_tmdb(name, config.tmdb_api_key)` directly. `classify_from_tmdb` makes blocking `requests.get` network calls, so it stalled the asyncio event loop for the duration of each TMDB lookup during disc identification — freezing WebSocket broadcasts, other concurrent jobs, and the drive sentinel.

## Change

- Made `_try_tmdb` an `async def` that offloads the lookup via `await asyncio.to_thread(classify_from_tmdb, name, config.tmdb_api_key)`.
- Awaited it at all three call sites: volume-label lookup, DINFO disc-name fallback, and the TMDB re-query after AI identification.
- The existing `try` / `except TmdbAuthError` / `except Exception` + `logger.warning(...)` handling is unchanged — `asyncio.to_thread` re-raises the worker thread's exception at the await point, so the handlers still catch TMDB failures.
- Mirrors the established sibling pattern in `_resolve_missing_tmdb_id`.

## Out of scope

- `classify_from_discdb` in the same method is also a blocking network call, but it's gated behind `DISCDB_ENABLED` (currently off) and was not flagged by the reviewer. Left as-is to keep this PR focused on the flagged TMDB path; can be a separate follow-up if DiscDB is re-enabled.
- No `CHANGELOG.md` entry: this matches #377, which made the analogous internal event-loop fix without a changelog entry. The behavior is unchanged; only event-loop responsiveness during identification improves.

## Verification

- `uv run pytest tests/integration/ -q` → **209 passed, 1 skipped**
- `uv run pytest tests/pipeline/test_classification.py tests/pipeline/test_ambiguous_movie_flow.py tests/pipeline/test_generic_label_flow.py -q` → **55 passed**
- `uv run ruff check app/services/identification_coordinator.py` → clean
- `uv run ruff format --check app/services/identification_coordinator.py` → already formatted

### DIFF
--- a/backend/app/services/identification_coordinator.py
+++ b/backend/app/services/identification_coordinator.py
@@ -1190,17 +1190,23 @@ class IdentificationCoordinator:
             None if config.tmdb_api_key else TMDB_DEGRADED_NOT_CONFIGURED
         )
 
-        def _try_tmdb(name: str, context: str):
+        async def _try_tmdb(name: str, context: str):
             """Run a TMDB lookup, swallowing and logging failures.
 
             ``context`` distinguishes the warning message between call sites.
             Returns the TMDB signal, or None on failure / no API key.
+
+            ``classify_from_tmdb`` makes blocking ``requests.get`` calls, so it is
+            offloaded to a thread to avoid stalling the event loop (mirrors
+            ``_resolve_missing_tmdb_id``). ``asyncio.to_thread`` re-raises the
+            worker thread's exception at the await point, so the handlers below
+            still catch TMDB failures.
             """
             nonlocal tmdb_degraded_reason
             if not config.tmdb_api_key:
                 return None
             try:
-                return classify_from_tmdb(name, config.tmdb_api_key)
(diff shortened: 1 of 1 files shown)


Jsakkos:
Thanks for the review. Addressed the `exc_info` nit in ad11af9: the generic `except Exception` branch in `_try_tmdb` now passes `exc_info=True`, matching its `TmdbAuthError` sibling, `_resolve_missing_tmdb_id`, and CLAUDE.md's error-handling guidance — a swallowed transient TMDB error now logs a full traceback. (No inline thread was created for this comment, so replying here.)

PR #431, opened 2026-06-21
test(discdb): remove inert try_discdb_assignment mocks after ASR-precedence refactor

Jsakkos:
## What

Removes leftover **inert** DiscDB test scaffolding in four unit test files. These mocks/patches still reference `try_discdb_assignment` on code paths that no longer touch it after the ASR-preferred precedence refactor on this branch (PR #430).

| File | Removed |
|------|---------|
| `test_identity_answer_midrip.py` | `try_discdb_assignment` patch in `_stub_matching` |
| `test_stuck_job_recovery.py` | 3 reconcile-path patches + a `fake_discdb` def + a `discdb.assert_not_called()` |
| `test_import_tmdb_resolution.py` | dead `coordinator._try_discdb_assignment` attribute assignment |
| `test_staging_import_state.py` | dead `coordinator._try_discdb_assignment` attribute assignment |

## Why they are inert (on this base)

After #430:
- `JobManager._dispatch_title_match` goes straight to `match_single_file` (no pre-dispatch DiscDB short-circuit).
- `IdentificationCoordinator.identify_from_staging` no longer calls an injected `_try_discdb_assignment` (the injection was dropped); `IdentificationCoordinator` has no such attribute, so the assignments in items 3 & 4 are dead dynamic attributes nothing reads.
- DiscDB episode assignment now runs **only** as a low-confidence fallback inside `MatchingCoordinator._match_single_file_inner`.

The tests in scope mock `match_single_file`/dispatch directly, so the fallback path never executes — the `try_discdb_assignment` mocks were never exercised.

## What is NOT touched

- The live `MatchingCoordinator.try_discdb_assignment` method and its fallback call in `_match_single_file_inner` are untouched.
- No production code changes; no change to what any test fundamentally verifies (TV import / identity / stuck-recovery assertions are all preserved).

## Base note

Stacked on the #430 branch (`claude/trusting-booth-d64e49`) **on purpose**: the production call-site removal that makes these mocks inert lives there, not on `main`. On `main` these mocks are still load-bearing (verified: removing one yields `TypeError: 'NoneType' object is not callable` at the live `identify_from_staging` call site). Merge after / on top of #430.

## Verification

```
cd backend && uv run pytest tests/unit/test_identity_answer_midrip.py tests/unit/test_stuck_job_recovery.py tests/unit/test_import_tmdb_resolution.py tests/unit/test_staging_import_state.py -q
# 64 passed
```
ruff clean on all four files.

### DIFF
--- a/backend/app/services/job_manager.py
+++ b/backend/app/services/job_manager.py
@@ -2835,15 +2835,17 @@ class JobManager:
         """Run the standard per-title match dispatch for a ripped file.
 
         Shared by ``_on_title_ripped``, ``reconcile_stuck_titles``, and
-        ``dispatch_pending_matches`` so the sites can't diverge: TheDiscDB
-        assignment first (then a completion check, since no match task will
-        run), otherwise a ``match_single_file`` task with the standard done
-        callback. Owns its session — match tasks must own their own sessions.
+        ``dispatch_pending_matches`` so the sites can't diverge: dispatches a
+        ``match_single_file`` task with the standard done callback. (DiscDB
+        episode assignment is no longer a pre-dispatch step; under ASR-preferred
+        precedence it runs only as a low-confidence fallback inside
+        ``_match_single_file_inner``.) Owns its session — match tasks must own
+        their own sessions.
 
         Skips a title that already has a live match task (see
         ``_inflight_match_dispatch``), making repeat dispatch safe. Returns
-        True when work was dispatched (DiscDB assignment or match task), False
-        when skipped (in-flight, or title vanished).
+        True when a match task was dispatched, False when skipped (in-flight,
+        or title vanished).
 
         TOCTOU note: the membership check and ``add`` are separated by NO
         awaits — both run in a single event-loop turn, so they are effectively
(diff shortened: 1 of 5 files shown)


Jsakkos:
Thanks! Addressed the stale-docstring note: `_dispatch_title_match`'s docstring intro now matches the implementation (dispatches `match_single_file`; DiscDB is a low-confidence fallback inside `_match_single_file_inner`), and the "DiscDB assignment or match task" return-value phrasing is corrected. Doc-only, no behavior change, pushed in 73c36fb. Put it on this branch (stacked on #430) rather than #430 directly since I can't push to that branch; merge ordering note still stands.

Jsakkos:
Rebased onto `main` and resolved the conflict. #430 has since squash-merged into `main` (commit `fca47d9`), so GitHub auto-retargeted this PR's base from the #430 branch to `main`. I replayed only my two commits with `git rebase --onto origin/main` (avoiding a replay of #430's already-squashed commits). Verified against `main`: the four test files still carry the inert mocks and the `_dispatch_title_match` docstring is still stale there, so both commits remain valid. `64 passed`, diff is the same 5 files.

The earlier **"do not merge before #430"** caveat is now moot — #430 is already in `main`.

_🤖 Updated by [Claude Code](https://claude.com/claude-code)_

PR #433, opened 2026-06-22
Add ARM64 (Linux aarch64) build for NVIDIA Jetson

Jsakkos:
## Why

A user wants to run Engram on an **NVIDIA Jetson** (aarch64 / Linux for Tegra). Today releases only ship Windows x64, Linux x86_64, and macOS arm64 bundles, so there's no artifact a Jetson can run. This adds a native `engram-linux-arm64.tar.gz`.

### Phase 1 — CPU-only arm64 release bundle (fully CI-tested)
- **`release.yml`**: new `build-linux-arm64` job on a GitHub-hosted `ubuntu-24.04-arm` runner. It mirrors the x86_64 job exactly — same frontend bundling, pinned PyInstaller, smoke test, CA-bundle / fpcalc / TLS self-test, and file manifest — only changing the arch assertion (`aarch64`) and artifact names. Wired into `create-release` (`needs`, artifact download, `sha256sums.txt`, and the release `files:` list).
- **`fetch_fpcalc.py`**: selects the fpcalc asset by `(system, machine)` instead of OS alone. Linux aarch64 uses **Chromaprint 1.6.0** (`chromaprint-fpcalc-1.6.0-linux-arm64.tar.gz`, SHA-pinned) because 1.5.1 never shipped an arm64 fpcalc; each asset entry now carries its own version so platforms bump independently. Verified the aarch64 binary is a valid `ARM aarch64` ELF containing `fpcalc`.
- **`updater.py`**: `_select_asset()` now distinguishes x86_64 from aarch64 so a Jetson downloads `engram-linux-arm64.tar.gz` and an x86_64 host stays on `engram-linux-x64.tar.gz`. New unit tests cover both.

### Phase 2 — On-device GPU ASR (documented + scripted; needs Jetson hardware to validate)
The PyPI `ctranslate2` aarch64 wheel is **CPU-only** — there is no CUDA aarch64 wheel — so a CI bundle (and the generic arm64 runner that builds it) can only produce a CPU build. GPU on Jetson requires compiling CTranslate2 from source against the device's JetPack CUDA toolkit.
- **`backend/scripts/jetson_gpu_setup.sh`**: builds CTranslate2 with `-DWITH_CUDA=ON -DWITH_CUDNN=ON -DWITH_MKL=OFF -DWITH_OPENBLAS=ON` for the target compute capability (Orin = 87, Xavier = 72), matching the bundle's CTranslate2 version, then swaps `libctranslate2.so*` + `ctranslate2/_ext*.so` into the bundle (backing up the CPU build).
- **`docs/development/jetson.md`**: install steps, the CPU/GPU split, prerequisites (JetPack 6.x), and how to confirm GPU is active. `gpu.md`, `README.md`, and `CHANGELOG.md` updated to reference the arm64 build and Jetson caveat.

## Testing
- `uv run pytest tests/unit/test_updater.py` — 55 passed, 2 skipped (incl. new arch-selection tests).
- `ruff check` + `ruff format --check` clean on changed files.
- `release.yml` YAML validated; `fetch_fpcalc.py` runs end-to-end on x86_64 (SHA verified) and resolves the correct key for aarch64/arm64.

## Status / caveats
- **Phase 1 is self-contained and shippable** (a runnable Jetson tarball with CPU ASR). It will be fully exercised by CI on the next release / a draft `workflow_dispatch` run.
- **Phase 2 (GPU) needs validation on real Jetson hardware** — it's signposted as best-effort in the script and docs. Device detection (`gpu_detected()`) may need a Tegra-specific check; flagged in the docs as a follow-up once hardware confirms.

https://claude.ai/code/session_01Vk2kqE9y5av9WJn6Bve5BB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vk2kqE9y5av9WJn6Bve5BB)_

### DIFF
--- a/backend/scripts/jetson_gpu_setup.sh
+++ b/backend/scripts/jetson_gpu_setup.sh
@@ -1,0 +1,153 @@
+#!/usr/bin/env bash
+#
+# jetson_gpu_setup.sh — enable CUDA-accelerated ASR for the Engram arm64 bundle
+# on an NVIDIA Jetson.
+#
+# WHY THIS EXISTS
+# ---------------
+# Engram transcribes episode audio with faster-whisper / CTranslate2. On x86_64
+# the PyPI `ctranslate2` wheel ships a CUDA-enabled build, so GPU "just works"
+# once cuDNN/cuBLAS are present. On aarch64 the PyPI wheel is **CPU-only** — there
+# is no CUDA aarch64 wheel — so the shipped `engram-linux-arm64.tar.gz` runs ASR
+# on the CPU. To get GPU on a Jetson you must compile CTranslate2 from source
+# against the device's JetPack CUDA toolkit and swap the result into the bundle.
+#
+# This script automates that. It is the on-device counterpart to the CPU-only CI
+# bundle, and it must be run ON THE JETSON (it compiles for the local GPU).
+#
+# IMPORTANT: this path requires validation on real Jetson hardware. Treat it as a
+# best-effort, well-signposted procedure rather than a turnkey guarantee.
+#
+# USAGE
+#   ./jetson_gpu_setup.sh /path/to/extracted/engram
+#
+#   where the argument is the extracted bundle directory that contains the
+#   `engram` launcher and the `_internal/` folder (i.e. the dir created by
+#   `tar xzf engram-linux-arm64.tar.gz`).
+#
+# OVERRIDES (environment variables)
+#   CUDA_ARCH   CUDA compute capability to target. Auto-detected when possible;
+#               defaults to 87 (Orin family). Use 72 for Xavier (AGX/NX/NANO),
+#               87 for Orin (AGX/NX/Nano), 53/62 for older TX/Nano.
+#   CT2_VERSION CTranslate2 git tag to build. Defaults to the version baked into
+#               the bundle (read from _internal/ctranslate2/version.py) so the
+#               compiled C++ library matches the bundled Python bindings.
+#   JOBS        Parallel build jobs (defaults to nproc).
+#
+set -euo pipefail
+
+err()  { printf '\033[31merror:\033[0m %s\n' "$*" >&2; exit 1; }
+info() { printf '\033[36m==>\033[0m %s\n' "$*"; }
+
+# --- 1. Validate arguments + host -------------------------------------------
+BUNDLE_DIR="${1:-}"
+[ -n "$BUNDLE_DIR" ] || err "usage: $0 /path/to/extracted/engram"
+BUNDLE_DIR="$(cd "$BUNDLE_DIR" && pwd)" || err "bundle dir not found: $1"
+CT2_PKG_DIR="$BUNDLE_DIR/_internal/ctranslate2"
+[ -d "$CT2_PKG_DIR" ] || err "not an Engram bundle (missing $CT2_PKG_DIR)"
+
+[ "$(uname -m)" = "aarch64" ] || err "this script must run on the Jetson (aarch64); got $(uname -m)"
+
+# A Jetson exposes its integrated GPU via the Tegra device nodes / L4T release
+# file. Bail early with a clear message rather than building something unusable.
+if [ ! -e /etc/nv_tegra_release ] && [ ! -e /dev/nvgpu ] && [ ! -e /dev/nvhost-gpu ]; then
+  err "no Jetson/Tegra GPU detected (no /etc/nv_tegra_release or /dev/nvhost-gpu).
+     This script is for NVIDIA Jetson with JetPack installed."
+fi
+
+# --- 2. Locate the JetPack CUDA toolkit -------------------------------------
+CUDA_HOME="${CUDA_HOME:-/usr/local/cuda}"
+[ -x "$CUDA_HOME/bin/nvcc" ] || err "CUDA toolkit not found at $CUDA_HOME (set CUDA_HOME).
+     Install it with JetPack:  sudo apt-get install nvidia-jetpack"
+info "CUDA toolkit: $("$CUDA_HOME/bin/nvcc" --version | sed -n 's/.*release \([0-9.]*\).*/\1/p' | tr -d '\n') at $CUDA_HOME"
+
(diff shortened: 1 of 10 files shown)


backend/scripts/jetson_gpu_setup.sh
```diff
@@ -0,0 +1,153 @@
...
+info "Compiling (this takes a while on a Jetson)..."
+cmake --build build --config Release -j "$JOBS"
+cmake --install build
+
+# --- 6. Build the matching Python bindings ----------------------------------
+# Build the bindings against the SAME CTranslate2 we just compiled so _ext.so
+# links the CUDA library. We target the bundle's interpreter (cp311) via a
+# disposable venv to avoid touching system Python.
+info "Building Python bindings..."
+python3 -m venv "$WORK/venv"
```
Jsakkos: **Python ABI mismatch: venv uses system `python3`, but the bundle is cp311.** The comment on line 107 claims this "targets the bundle's interpreter (cp311)", but `python3 -m venv` uses whatever `python3` resolves to — and JetPack 6 (Ubuntu 22.04) ships **Python 3.10** by default. The bindings then build a `_ext.cpython-310-...so`, which the frozen PyInstaller runtime (cp311, with its own bundled libpython3.11) cannot import — so the swap at line 133 produces an `ImportError` instead of GPU ASR.

The script should pin/verify a 3.11 interpreter and fail loudly if it's absent, e.g.:

```suggestion
PY311="${PYTHON:-python3.11}"
command -v "$PY311" >/dev/null || err "python3.11 is required to match the bundle's cp311 ABI (got $($PY311 --version 2>&1 || echo none)). Install it (e.g. sudo apt-get install python3.11 python3.11-venv) or set PYTHON."
"$PY311" -m venv "$WORK/venv"
```

---
_Generated by [Claude Code](https://claude.ai/code)_

backend/scripts/jetson_gpu_setup.sh
```diff
@@ -0,0 +1,153 @@
...
+# Replace the CPU-only ctranslate2 in the bundle with the freshly built CUDA one:
+#   - libctranslate2.so*           -> _internal/ (next to the loader's rpath)
+#   - ctranslate2/_ext*.so + *.py  -> _internal/ctranslate2/
+info "Backing up the CPU-only ctranslate2 ($CT2_PKG_DIR.cpu-backup)..."
+rm -rf "$CT2_PKG_DIR.cpu-backup"
+cp -a "$CT2_PKG_DIR" "$CT2_PKG_DIR.cpu-backup"
+
+UNPACK="$WORK/unpack"
+mkdir -p "$UNPACK"
+( cd "$UNPACK" && unzip -q "$WORK"/wheel/ctranslate2-*.whl )
```
Jsakkos: **`unzip` is used but never installed.** The apt step (lines 81–82) installs `git build-essential cmake libopenblas-dev` but not `unzip`, and a fresh Jetson/L4T image doesn't ship it. With `set -euo pipefail`, this line aborts the script right after the long CTranslate2 compile. Either add `unzip` to the apt list, or avoid the dependency entirely with the stdlib:

```suggestion
( cd "$UNPACK" && python3 -m zipfile -e "$(echo "$WORK"/wheel/ctranslate2-*.whl)" . )
```

---
_Generated by [Claude Code](https://claude.ai/code)_

backend/scripts/jetson_gpu_setup.sh
```diff
@@ -0,0 +1,153 @@
...
+
+info "Installing CUDA ctranslate2 into the bundle..."
+# Python package files (incl. the compiled _ext*.so).
+cp -a "$UNPACK"/ctranslate2/. "$CT2_PKG_DIR"/
+# The shared library: prefer the one the wheel bundles; fall back to the install tree.
+SO="$(find "$UNPACK" "$WORK/ct2-install" -name 'libctranslate2.so*' | head -n1)"
+[ -n "$SO" ] || err "could not locate built libctranslate2.so"
+cp -a "$SO"* "$BUNDLE_DIR/_internal/" 2>/dev/null || cp -a "$SO" "$BUNDLE_DIR/_internal/"
```
Jsakkos: **Copying only the first `libctranslate2.so*` match can drop the SONAME symlink the loader needs.** `find ... | head -n1` picks one arbitrary path; if it returns the versioned real file (e.g. `libctranslate2.so.4.6.3`), then `cp -a "$SO"*` matches only that file and any longer suffix — it won't copy the `libctranslate2.so.4` symlink. CTranslate2's `_ext` is linked against the SONAME (`libctranslate2.so.4`), so `dlopen` fails at import with "cannot open shared object file" even though a `.so` is present. Copy the whole set of `libctranslate2.so*` from the containing directory (preserving symlinks) rather than a single `head -n1` match.

---
_Generated by [Claude Code](https://claude.ai/code)_

Jsakkos (commented):
Automated code review (the `Code Review` workflow was skipped for this PR — it triggers only on `pull_request: opened` and was guarded out because the PR opened as a draft; `ready_for_review` isn't in its trigger list, so it never re-ran). I ran the equivalent review on the diff.

**CI-tested core is clean.** `release.yml` (the new `build-linux-arm64` job), `fetch_fpcalc.py` (arch-aware asset selection), and `updater.py` (`_select_asset` x86_64-vs-aarch64) hold up — the arch selection logic is correct and covered by unit tests, and the bundle is gated by the existing smoke / fpcalc / TLS self-test.

**3 findings, all in the on-device GPU helper** `backend/scripts/jetson_gpu_setup.sh` (inline below). It's documented as best-effort/needs-hardware-validation, but these would each cause a real failure on the first Jetson run:
1. venv built with system `python3` (3.10 on JetPack 6) produces a cp310 `_ext.so` that the cp311 frozen runtime can't import — contradicts the line-107 comment.
2. `unzip` is invoked but never installed (and absent on a fresh L4T image).
3. copying only the first `libctranslate2.so*` match can omit the SONAME symlink the loader needs.

None block Phase 1 (the shippable CPU arm64 tarball); they're worth fixing before the GPU path is validated on hardware.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #466, opened 2026-06-30
fix(ui): replace window.confirm cancel with Synapse v2 modal

katelovescode:
Single-click cancel was an accidental-tap risk (ask me how I know 😭). This adds an animated modal matching the Synapse v2 design language (SvPanel, mono uppercase, red glow on confirm). Adds two unit tests covering modal open/dismiss and confirm paths.

PR #479, opened 2026-07-01
fix(ui): focus management in ConfirmModal (WCAG 2.4.3)

katelovescode:
## Summary

- On modal open: captures `document.activeElement` (the trigger button), then focuses the dismiss button — the safer default per ARIA dialog pattern
- On modal close (both confirm and dismiss paths): restores focus to the captured trigger via the `useEffect` cleanup

Self-contained in `ConfirmModal` — no caller changes needed.

## Context

Addresses the focus management gap noted in the re-review of #466:
> When the modal opens, focus remains on the button that triggered it. When it closes, focus stays wherever it landed. WCAG 2.4.3 (Focus Order) and the ARIA dialog pattern both expect focus to move into the dialog on open, and return to the trigger on close.

## Test plan

- [ ] Open cancel modal → focus lands on "Keep Ripping" (dismiss) button
- [ ] Dismiss with keyboard (Enter/Space on dismiss, or Escape) → focus returns to Cancel button on the card
- [ ] Confirm cancel → focus returns to Cancel button (before card disappears)
- [ ] Same for Force-advance modal

### DIFF
--- a/frontend/src/app/components/DiscCard/ActionButtons.tsx
+++ b/frontend/src/app/components/DiscCard/ActionButtons.tsx
@@ -17,7 +17,7 @@
  * promote it to the primary callout in that state.
  */
 
-import { useState, useEffect, type CSSProperties, type ReactNode, type MouseEvent } from "react";
+import { useState, useEffect, useRef, type CSSProperties, type ReactNode, type MouseEvent } from "react";
 import { createPortal } from "react-dom";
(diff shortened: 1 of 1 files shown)


katelovescode:
Ready for review now after fixing the prior run

Jsakkos:
@claude please review this pr

Issue #459, opened 2026-06-27
[Bug] Job 2 failed in failed


MadsTheEngineer:
## Bug Report

**Engram version**: 0.21.12
**OS**: Linux 6.8.0-117-generic
**Python**: 3.11.15
**Build**: dev (sys.frozen=False, _MEIPASS=False)
**MakeMKV**: MakeMKV v1.18.4 linux(x64-release)
**FFmpeg**: ffmpeg version 5.1.9-0+deb12u1 Copyright (c) 2000-2026 the FFmpeg developers

### Job Context
- **ID**: 2
- **Label**: 
- **Type**: unknown
- **State**: failed
- **Error**: No titles found on disc

### Steps to Reproduce
1. Insert 1995 Pride and Prejustice DVD 1 into drive

### Expected Behavior
1. Would start ripping

### Actual Behavior
1. Error after 1 second. Cannot eject disc. Cannot restart container. Has to reboot system.
2. log: docker compose restart
[+] Restarting 0/1
 ⠙ Container engram  Restarting                                                                                                            14.2s 
Error response from daemon: Cannot restart container 0c69285d48e333b50bc81ad5dae94992a8a6e6116fc74ff90eaf4061a06808eb: permission denied

Jsakkos:
@MadsTheEngineer Have you had any other successful jobs? The docker images are 100% untested by me.

MadsTheEngineer:
> [@MadsTheEngineer](https://github.com/MadsTheEngineer) Have you had any other successful jobs? The docker images are 100% untested by me.

Only partly. 
I have tried with Breaking Bad. Other identified is as Breaking Bad Cereal (2014) and or only ripped one file. I does not show up in the History, it did not show when running and it is not en either review or the success column.

MadsTheEngineer:
I also tried a movie (Zombieland (2009̀)) it seems it was identified correctly and ripped, but it did not show during up during ripping. Actually I cannot see any trace of it inside Engram, but it showed up in the output folder for movies.

Jsakkos:
Can you share some of your logs? If you can find one of the jobs in the history tab, click on it, then scroll to the bottom and click "report bug for this job.

MadsTheEngineer:
I see this in the scan logs: 

MSG:1005,0,1,"MakeMKV v1.18.4 linux(x64-release) started","%1 started","MakeMKV v1.18.4 linux(x64-release)"
[...]

PR #481, opened 2026-07-01
fix(config): expand ~ before absolute-path check in ensure_paths_exist

Jsakkos:
## Summary
- `ensure_paths_exist()` checked `Path.is_absolute()` before calling `.expanduser()`, but `Path("~/foo").is_absolute()` is always `False` in pathlib, so the default `subtitles_cache_path` (`~/.engram/cache`) was never tilde-expanded. It instead resolved to a literal `~` directory joined onto the backend source dir (e.g. `/app/app/~/.engram/cache` in Docker), which is unwritable for the unprivileged runtime user — producing a permission-denied warning on every startup/config save.
- Fixes by calling `.expanduser()` first, matching every other consumer of `subtitles_cache_path` (`curator.py`, `precomputed_cache_service.py`, `routes.py`, etc.).
- Found while triaging [#459](https://github.com/Jsakkos/engram/issues/459) — this bug is cosmetic there (staging/library paths were already absolute for that user), not the cause of the reported ripping issue, but worth fixing since it's a real defect.

## Test plan
- [x] Added `test_ensure_paths_exist_expands_tilde` regression test (fails on the old code, passes after the fix)
- [x] `uv run pytest tests/unit/test_config_service.py tests/unit/test_setup_gate.py` — 20 passed
- [x] `uv run ruff check` / `uv run ruff format --check` on changed files — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Engram will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Jsakkos:
@claude please review this PR

PR #483, opened 2026-07-02
fix(db): self-heal Alembic upgrades stuck behind a duplicate column

Jsakkos:
## Summary
- `_add_missing_columns()` (the frozen-build fallback) and Alembic both run on every startup and converge the schema toward the same SQLModel metadata. When `_add_missing_columns()` won the race and added a column first, Alembic's own migration for that same column failed with `duplicate column name` and was silently swallowed (`logger.warning(..., "non-fatal")`) — leaving `alembic_version` stuck one revision behind forever and silently blocking every later migration on every subsequent restart.
- Surfaced by a user's real-world startup log in [#459](https://github.com/Jsakkos/engram/issues/459) (`duplicate column name: import_manifest_json`), separate from the tilde-expansion bug already fixed there in #481.
- `_run_alembic_upgrade()` now applies pending migrations one revision at a time via a new `_upgrade_to_head_self_healing()`, and stamps past any single revision whose only failure is a column that already exists, instead of getting stuck.
- Note: `#459` itself remains open — this fixes a newly-found migration bug from the thread, not the originally reported "disc doesn't rip / jobs invisible in UI" symptom, which is still waiting on a requested debug log from the reporter.

## Test plan
- [x] Added `test_upgrade_heals_when_head_column_already_present` — reproduces the exact `duplicate column name` failure against the real Alembic migration chain, confirmed it fails on the old code before implementing the fix
- [x] `uv run pytest tests/unit/ tests/pipeline/` — 2511 passed, 19 skipped
- [x] `uv run ruff check` / `uv run ruff format --check` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to Engram will be documented in this file.
 ### Fixed
 
 - **Subtitle cache directory no longer fails to create with a permission-denied warning on startup.** `ensure_paths_exist()` checked `is_absolute()` before expanding `~`, so the default `subtitles_cache_path` of `~/.engram/cache` was never tilde-expanded and instead resolved to a literal `~` folder under the backend source directory, which is unwritable in Docker. `~` is now expanded before the absolute-path check, matching every other consumer of this setting. (#459)
+- **Database migrations no longer get stuck behind a "duplicate column name" error.** `_add_missing_columns()` (the frozen-build fallback) and Alembic both converge the schema toward the same model on every startup; if a column a pending Alembic migration wanted to add had already been added out-of-band, the migration failed and was silently swallowed, leaving `alembic_version` permanently one revision behind and blocking every later migration on every subsequent restart. Alembic upgrades now apply one revision at a time and self-heal past a revision whose only failure is a column that already exists. (#459)
 
 ## [0.22.2] - 2026-06-30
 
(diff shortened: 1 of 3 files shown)


Jsakkos:
@claude please review this PR

Jsakkos:
Addressed both points:

1. **Hardcoded `max_steps=25`** — replaced with a dynamic bound derived from `len(list(script.walk_revisions())) + 1`, so it no longer needs manual bumping as migrations accumulate. Validated this working as intended: while this fix was in review, `main` picked up a new migration (`6148bcd5c13a_add_app_config_discord_webhook_url`) via a routine branch update/merge, and the self-heal step count adjusted automatically with no code change needed.
2. **Undocumented invariant** — added a docstring note on `_upgrade_to_head_self_healing` spelling out that a migration must not combine `add_column()` with other DDL/data ops in the same `upgrade()`, since the self-heal path assumes the whole revision is a no-op when it hits a duplicate-column error.

Also added `test_self_heal_reraises_non_duplicate_column_errors` per your suggestion, confirming a genuinely different `OperationalError` still propagates rather than being silently swallowed.

Left the "two consecutive duplicate-column revisions" test and the outer-`except Exception` swallowing as-is — the former would need a synthetic migration chain to construct cleanly against the real chain, and the latter intentionally preserves the pre-existing fail-open startup behavior (a `RuntimeError` from exhausted retries degrading to the same non-fatal warning as any other Alembic failure, rather than crashing the app).

Full suite: 2523 passed, 19 skipped after merging in `main`.

PR #502, opened 2026-07-12
feat: add Background Animation toggle for low-power devices

Jsakkos:
## Summary
- Adds a **Background Animation** toggle in Settings → Preferences → Display that disables the falling-code rip animation (`SvRipAnimation`) — a full-viewport canvas redrawn at 20fps that's a meaningful CPU/GPU cost on low-power devices (reported case: ARM64). Stored in `localStorage`, not backend `AppConfig`, since it describes the browser rendering the dashboard rather than the backend host. Defaults on (backward compatible), seeded from `prefers-reduced-motion` when no explicit choice has been made.
- Two bugs were found and fixed during manual browser verification (not caught by the jsdom test suite, which runs under `MotionConfig reducedMotion="always"`):
  - The hook's `useState` was per-component-instance, so toggling the settings checkbox didn't propagate to `App.tsx`'s already-mounted gate without a full reload. Fixed with a same-tab custom `window` event broadcast so every mounted instance re-syncs immediately.
  - A pre-existing issue in `SvAtmosphere.tsx` (unrelated to this feature, but newly reachable through it): `AnimatePresence`'s exit-then-remove bookkeeping didn't reliably fire in real browsers (confirmed in both `npm run dev` and a production build), leaving the animation layer permanently stuck at `opacity:0`, attached but frozen, unable to resume. Fixed by dropping `AnimatePresence` for this layer in favor of a plain conditional mount — the fade-in still works, only the fade-out transition is lost in exchange for reliable mount/unmount.

## Test plan
- [x] `npx vitest run` — 313/313 tests pass across 35 files
- [x] `npx eslint` / `npx tsc --noEmit` — clean
- [x] Manual verification in both `npm run dev` and a production build (`npm run build && vite preview`): toggle off stops the canvas redraw loop and removes the DOM node; toggle on resumes it; preference persists across reload and syncs live across mounted instances without a reload
- [ ] Reviewer: exercise the toggle in Settings → Preferences → Display with a simulated ripping disc (`POST /api/simulate/insert-disc`, `DEBUG=true`)

Design spec: `docs/superpowers/specs/2026-07-09-background-effects-toggle-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-09-background-effects-toggle.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Engram will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Background Animation toggle.** A new Settings → Preferences → Display checkbox disables the falling-code rip animation, a full-viewport canvas redrawn at 20fps that's a meaningful CPU/GPU cost on low-power devices (reported case: ARM64). Stored per-browser (not synced across devices viewing the same backend), defaults on, and seeds from the OS's reduced-motion setting when no explicit choice has been made. (#502)
+
 ## [0.24.0] - 2026-07-04
 
 _Highlights: manual track skipping for queued/not-yet-ripped tracks, plus field-validated Jetson GPU setup fixes._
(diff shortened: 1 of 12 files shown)


Jsakkos:
@claude please review this PR

Jsakkos:
Addressed the changelog nit: added an `[Unreleased]` entry for the Background Animation toggle (b366acc3).

PR #503, opened 2026-07-12
fix(analyst): apply uncorroborated-identity review gate to movies too

### LINKED ISSUE #499: [Bug] Miss Identifed TV Show As Movie
a TV series is getting matched as a movie, it rips all 3 episodes, then just deletes the files. I have no option to reidentify back as a TV Show

[engram-bug-report-job-103.zip](https://github.com/user-attachments/files/29905922/engram-bug-report-job-103.zip)

Jsakkos:
## Summary

Fixes #499. Root-caused via the diagnostics bundle attached to the issue (job 103):

- The disc has only **2** real episode-length titles (43min, 43min) plus a Play-All-length title (84min). TV cluster detection in `_detect_tv_show` requires **3+** similarly-sized titles, so this disc can never be recognized as TV by duration alone, and the label (`DH20NNT5`) has no season pattern to trigger the label-fallback path either.
- With no TV signal, `_detect_movie` wins by default (the 84min title alone clears the 80-min movie floor).
- DiscDB/TMDB lookup on the opaque label failed, so identification fell back to AI, which guessed an unrelated, low-popularity movie ("The New Mutants").
- The analyst *does* have a check for exactly this — "TMDB name corroborated by neither the volume-label name nor the disc title" — but the review-escalation that's supposed to follow (`identity_unconfirmed`) was explicitly scoped to `content_type == TV` only ("movie discs keep their prior auto-accept behavior").
- Because the heuristic and the wrong TMDB guess both said "movie," that agreement boosted confidence and cleared the auto-rip gate (`needs_review` stayed `False`). The job ripped and auto-organized, filing the two real episodes as `Extras/Extra 1.mkv` / `Extra 2.mkv` with no way back — `re-identify` only works on `RIPPING`/`REVIEW_NEEDED` jobs, and this one went straight to `COMPLETED`.

## What changed

Removed the TV-only restriction on the uncorroborated-identity review gate in `backend/app/core/analyst.py` (two call sites: the heuristic+TMDB agreement path, and the TMDB-only path). An uncorroborated movie identity now gets the same treatment TV already has — it rips first (walk-away preserved) but parks with a confirm/correct prompt via the existing `re-identify` endpoint and `ReIdentifyModal` (which already supports switching movie↔tv), instead of silently completing.

This is the narrowly-scoped fix (discussed as "Fix A" in the issue thread) — it does not change the underlying 2-episode TV-cluster detection gap or add a reclassify path for already-`COMPLETED` jobs; those were explicitly left out of scope.

## Test plan

- [x] New test `test_uncorroborated_movie_identity_sets_flag` reproduces job 103's exact disc shape (43/43/84 min, label `DH20NNT5`, TMDB movie match "The New Mutants") — written first, watched it fail (`needs_review` stayed `False`), then the fix made it pass.
- [x] Updated one pre-existing test (`test_tmdb_resolves_unknown_to_movie`) whose assertion encoded the same bug on a different code path (uncorroborated "MYSTERY_DISC" → "Inception") — now asserts the corrected behavior.
- [x] `uv run pytest tests/unit` — 2436 passed, 19 skipped
- [x] `uv run pytest tests/pipeline` — 99 passed
- [x] `uv run pytest tests/integration tests/real_data` — 253 passed, 8 skipped, 1 failed (pre-existing, unrelated: a real-optical-drive hardware test that fails identically with the change stashed out)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to Engram will be documented in this file.
 
 ### Fixed
 
+- **An unconfirmed movie identity no longer auto-accepts.** When a disc's title couldn't be confirmed by DiscDB/TMDB from its label, an AI-guessed title was sometimes matched to an unrelated, low-popularity movie — and because that guess "agreed" with the movie heuristic, the mismatch was silently accepted and ripped straight through, misfiling real content as movie extras with no way to correct it afterward (#499). The disc-identity corroboration check that already gated this for TV shows now applies to movies too: an uncorroborated TMDB/AI identity pauses for confirmation via the existing re-identify prompt instead of completing unreviewed.
 - **The Import folder browser now scrolls.** Browsing a directory with more subfolders than fit on screen previously grew the modal past the edges of the window, stranding you in the middle of the list with no way to scroll, no visible way to go up a level, and no way to reach the start button. The folder list is now a scrolling pane inside a window-sized modal, so the header, the up-a-level row, and the start button stay put. You can also type or paste an exact folder path instead of clicking through hundreds of folders, and going up a level scrolls the folder you came from back into view.
 
 ## [0.24.0] - 2026-07-04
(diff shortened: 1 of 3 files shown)


Jsakkos:
@claude please review this PR

PR #508, opened 2026-07-14
feat: customizable Discord notification templates

### LINKED ISSUE #507: feat: customizable Discord notification templates — implementation approach input needed
## Summary

Discord notifications currently use a hard-coded message format (`backend/app/core/discord_notifier.py`). We want to let users customize notification content via templates with variables like `{{title}}`, `{{drive}}`, `{{duration}}`, `{{error}}`, etc.

Scope for this change:
- Applies to the two existing notification triggers only: job `COMPLETED` and `FAILED` (`job_manager.py:1067`, `job_state_machine.py` terminal callbacks).
- Variable set will be fairly comprehensive — most `DiscJob` fields relevant to a completed/failed job (title, drive, content type, TMDB metadata, timestamps/duration, error message, subtitle status, etc.).

## Current implementation

- `discord_notifier.py:7` — `notify_discord(webhook_url, job_id, label, state)` builds a fixed embed (title/description/color only) and POSTs via `httpx`.
- `job_manager.py:1067-1094` — `_notify_discord_on_terminal` / `_send_discord_notification` wire this to the state machine's terminal callback, fire-and-forget via `asyncio.create_task`.
- Config: single `AppConfig.discord_webhook_url` string, SSRF-validated on save.

## Question about implementation

Since templates would be user-authored (via config/API) and evaluated server-side, there are security and capability tradeoffs to consider. Three options considered:

1. **chevron (Mustache implementation)** — `{{var}}` syntax matches modern common templating patterns. No arbitrary expression evaluation, with conditional sections (`{{#error}}...{{/error}}`) to cleanly omit fields that don't apply to a given job. Pure-Python, small, actively maintained.

2. **Jinja2** — built in conditionals, filters, loops

3. **stdlib `string.Template`** (subclassed for `{{var}}` delimiters): plain substitution only, no way to conditionally omit a line for a field that's empty/inapplicable (e.g. no error line on a successful rip) without pre-processing logic outside the template.

Let me know what you think @Jsakkos and I'll write a PR.

katelovescode:
## Summary

Closes #507. Discord embeds sent on job COMPLETED/FAILED now render from a user-editable template instead of a fixed message.

- Two new config fields — `discord_template_completed` / `discord_template_failed` — using **chevron** (`{{var}}` mustache syntax), per your comment on #507.
- Template controls the embed **description text only**; title/color/emoji (✅ green / ❌ red) stay hardcoded.
- Validated at `PUT /api/config` (422 on unknown variable or malformed syntax) so a bad template can never reach a live job — never validated/rejected at render time.
- **Live validation in ConfigWizard**: a new `POST /api/validate/discord-template` endpoint reuses the same server-side validator (no second parser reimplemented in TypeScript). Each template textarea debounces (400ms) a call to it, shows the error inline, and disables Save Changes / Next while either template is currently invalid — catching mistakes before Save instead of after a 422 toast. This is UX only; the real enforcement stays at `PUT /api/config`, so it can't be used to bypass validation.
- Available variables: `title`, `drive`, `job_id`, `content_type`, `season`, `tmdb_name`, `tmdb_year`, `duration` (derived from `completed_at - created_at`), `error`, `subtitle_status`, `subtitles_downloaded`, `subtitles_total`, `subtitles_failed`, `path`, `total_titles`.
- Empty template string = use the built-in default (`**{{title}}**`, matching current behavior byte-for-byte).
- New Alembic migration adds the two `app_config` columns.
- ConfigWizard gets two new textareas in the Notifications section.

## Design notes / open questions for review

- **Status code**: template validation returns 422, while the existing naming-format check in the same endpoint returns 400 — pre-existing inconsistency, not touched here.
- `notify_discord()` now takes a pre-rendered `description: str` rather than doing the templating itself — keeps chevron out of the low-level POST helper's hot path/test surface. Happy to restructure if you'd rather see a `discord_templates.py` split.
- `{{drive}}` kept separate from `{{title}}` despite overlap in the common case (title falls back to drive label) — lets a template reference the raw label even when a TMDB override exists.
- Introduced `TERMINAL_JOB_STATES` (`app/models/disc_job.py`) to replace four separate `(JobState.COMPLETED, JobState.FAILED)` literal tuples scattered across `job_state_machine.py` and `job_manager.py` — one of them was added by this PR (a defensive guard in `_send_discord_notification`), which is what surfaced the existing duplication. Small refactor, no behavior change, but touches code outside the Discord feature itself — flagging in case you'd rather see it split out.

## Test plan

- [x] Backend: `uv run pytest` (full suite green) — new unit tests for `validate_discord_template`, `build_template_context`, `render_discord_template`, updated `notify_discord`/`_send_discord_notification` tests, new `PUT /api/config` validation tests (422/200/blank-ok), new `POST /api/validate/discord-template` endpoint tests.
- [x] `uv run ruff check .` / `uv run ruff format --check .`
- [x] Frontend: `npx vitest run` (full suite green) — new ConfigWizard tests for template load/save roundtrip, 422 toast surfacing, live-validation error display, and Save/Next disabling; new `discordTemplateValidation` util tests.
- [x] `npm run build` (typecheck) / `npm run lint`
- [x] Migration manually verified against a simulated pre-migration DB (add_column path), not just autogenerate diff.
- [x] Manually confirmed via local dev server: live validation errors and re-enabling of Save Changes work end-to-end.
- [x] Manual: set a webhook + custom template via the UI, trigger a simulated COMPLETED/FAILED job, confirm the rendered Discord message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Engram will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Customizable Discord notification messages.** The Discord embed sent when a disc completes or fails now uses a template you control, instead of a fixed message. Set separate messages for the completed and failed cases in Settings → Notifications, using `{{title}}`, `{{drive}}`, `{{duration}}`, `{{error}}`, and other job details as variables. Leave either field blank to keep the previous default message.
+
 ### Fixed
 
 - **Docker deployments now run `privileged: true` by default.** Plain `devices:` passthrough wasn't always enough for full optical drive control (eject/tray), and some hosts got stuck unable to stop or restart the container without it. (#459, thanks @MadsTheEngineer!)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,6 +348,7 @@ Components use updated settings
 - **API Keys**: `makemkv_key`, `tmdb_api_key` (redacted in responses)
 - **Matching**: `max_concurrent_matches` (default: 3), threshold constants in Analyst
 - **Conflict resolution**: `conflict_resolution_default` ("skip" | "overwrite" | "ask")
+- **Discord notifications**: `discord_template_completed` / `discord_template_failed` — customizable embed description (chevron `{{var}}` mustache syntax, see `app/core/discord_notifier.py::ALLOWED_TEMPLATE_VARS`); empty string = built-in default
 
 ### Configuration Validation
 
--- a/backend/app/api/routes.py
+++ b/backend/app/api/routes.py
@@ -374,6 +374,8 @@ class ConfigResponse(BaseModel):
     contribution_pseudonym: str | None = None
     # Notifications
     discord_webhook_url: str = ""
+    discord_template_completed: str = ""
+    discord_template_failed: str = ""
 
 
 class ConfigUpdate(BaseModel):
@@ -459,6 +461,8 @@ class ConfigUpdate(BaseModel):
     fingerprint_disclosure_accepted: bool | None = None
     # Notifications
     discord_webhook_url: str | None = None
+    discord_template_completed: str | None = None
+    discord_template_failed: str | None = None
 
 
 class ReviewRequest(BaseModel):
@@ -1533,6 +1537,8 @@ async def get_config() -> ConfigResponse:
         contribution_pseudonym=config.contribution_pseudonym,
         # Notifications
         discord_webhook_url="***" if config.discord_webhook_url else "",
+        discord_template_completed=config.discord_template_completed,
+        discord_template_failed=config.discord_template_failed,
     )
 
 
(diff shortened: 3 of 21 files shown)


Jsakkos:
@claude please review this PR.

katelovescode:
I'll address these, they actually don't seem like nits to me

katelovescode:
Re: comment-tag nit (`{{! note }}` flagged as unknown variable): couldn't reproduce.

Chevron's `tokenize()` filters `tag_type="comment"` internally and never yields it, so `validate_discord_template("{{title}} {{! a comment }}")` already returns `None` (valid) on this branch. Verified directly against `tokenize()` output and the validator itself.

Added a regression test (`test_validate_discord_template_accepts_comment_tag`) pinning this behavior, in case a future chevron version changes it. No code fix needed here.

Fixed in 19fec14:
- Default templates now use unescaped `{{{title}}}` instead of `{{title}}`, so titles with `&`/`<`/`>`/`"` render correctly instead of HTML-escaped.
- `render_discord_template` now passes `partials_dict={}`, closing the `{{>title}}` partial-tag file-read edge case.
- ConfigWizard's live-validation debounce now skips the `/api/validate/discord-template` round-trip when a template field is empty (always valid — falls back to the default), fixing the unnecessary calls on mount and while fields are blank.

katelovescode:
@Jsakkos ready for re-review

Jsakkos:
@claude please review this PR again.

PR #536, opened 2026-07-22
fix: allow manual import from LAN when allow_lan_access is enabled (#524)

### LINKED ISSUE #524: Import feature not working in Docker
### Discussed in https://github.com/Jsakkos/engram/discussions/521

<div type='discussions-op-text'>

<sup>Originally posted by **mannaoverquail** July 18, 2026</sup>
I am running this in a docker container (on a different machine than I am accessing on) but whenever I try to use the import feature, I get this error: Request failed (403 Forbidden): {"detail":"This endpoint is only reachable from the host machine"}

I tried some reverse proxy stuff, but that didn't work. I don't know if that is even the right path to go down. Anyone have any ideas? </div>

Jsakkos:
Fixes #524.

## Problem

Manual import failed in Docker/remote deployments with `403 Forbidden: {"detail":"This endpoint is only reachable from the host machine"}`. The import endpoints (`/import/browse`, `/import/preview`, `/import/start`) were guarded by `require_localhost`, which rejects **any** non-loopback client unconditionally. When the dashboard is accessed from another machine (the whole point of a headless/Docker install), the request never originates from `127.0.0.1`, so import was architecturally unreachable — even though headless first-run already seeds `allow_lan_access=True`.

## Fix

Added `require_localhost_or_lan`: loopback is always allowed, and LAN peers are permitted **only when the user has enabled `allow_lan_access`**. The three import endpoints now use this guard; the more sensitive fingerprint/contribution endpoints keep the strict `require_localhost`. Config-read failures fail closed (deny).

The 403 message now points users to the LAN access setting. The frontend surfaces the response body verbatim, so the improved guidance reaches the UI with no frontend change.

## Tests

- New `TestRequireLocalhostOrLan`: loopback-always-allowed (never reads config), LAN-rejected-when-off, LAN-allowed-when-on, fail-closed-on-config-error.
- `tests/integration/test_import_endpoints.py`: 12 passed. Ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Engram will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The manual import feature now works in Docker and other remote setups.** Browsing for files and starting an import over the network returned `403 Forbidden` because those endpoints were locked to the host machine. They now honor the **Allow LAN access** setting: when you've enabled it (headless/Docker installs enable it automatically on first run), import is reachable from another device, while the more sensitive fingerprint and contribution endpoints stay host-only. (#524)
+
 ## [0.26.0] - 2026-07-22
 
 _Highlights: enter your own disc details when Engram can't work them out, plus customizable Discord notification messages._
--- a/backend/app/api/routes.py
+++ b/backend/app/api/routes.py
@@ -114,6 +114,40 @@ def require_localhost(request: Request) -> None:
         )
 
 
+async def require_localhost_or_lan(request: Request) -> None:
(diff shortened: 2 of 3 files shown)


Jsakkos:
@claude please review this PR

Jsakkos:
Addressed the fixture nit: the `client` fixture in `test_import_endpoints.py` now overrides `require_localhost_or_lan` (the guard the import routes actually use) instead of the now-unused `require_localhost`. The LAN/loopback/fail-closed branches remain covered directly in `TestRequireLocalhostOrLan`. Pushed in e59995ba; 12 passed.

PR #537, opened 2026-07-22
fix: prevent /api/config 500 on DBs upgraded to 0.26.0

Jsakkos:
## Summary

Fixes a **0.26.0 upgrade regression** that returns HTTP 500 from `GET /api/config` on any database created before the customizable Discord messages feature (#508). Because config loading powers the whole dashboard, the first launch of 0.26.0 against a pre-existing DB broke Settings, the Contribute page, and the first-run wizard.

Reproduced against a real pre-0.26.0 `~/.engram/engram.db`.

## Root cause

The new `discord_template_completed` / `discord_template_failed` columns (`app_config`, model default `str = ""`) landed as **NULL** on upgrade. `ConfigResponse` types them as required `str`, so a `None` raised a pydantic `ValidationError` → 500.

The mechanism was subtler than "column defaults to NULL": these columns are `nullable=False`. The real bug is that SQLModel types `str` fields as **`AutoString`**, a `TypeDecorator` that is **not** a subclass of `sqlalchemy.types.String`. So `_add_missing_columns`' `isinstance(col.type, String)` check missed it and fell through to the no-default branch, and `ALTER TABLE ADD COLUMN` left existing rows NULL. (Frozen/dev-fallback builds skip Alembic, so this reconciler is the path that runs.)

## Fix

- **`database.py`** — new `_model_default_literal(col)` renders a column's declared scalar Python default as a SQL literal; `_add_missing_columns` prefers it over the type-based fallback. This backfills existing rows with the model's declared default for **all** column types, not just these two.
- **`routes.py`** — defensive `or ""` coalesce on the two Discord fields in `get_config`, covering DBs an *early* 0.26.0 build already NULLed (where re-running `_add_missing_columns` won't re-add the existing column).

## Tests

- `test_add_missing_string_column_backfills_declared_default` (migration layer) — asserts the reconciler backfills `''`, not NULL.
- `test_get_config_survives_null_discord_templates` (route layer) — reproduces the real nullable-column shape and asserts `GET /api/config` returns 200.

Both confirmed to fail before the fix. Full unit suite: `2539 passed, 19 skipped`; `ruff check` clean.

## Reviewer notes

- Behavior change: `_add_missing_columns` now backfills declared defaults instead of NULL for **every** added column (e.g. `fingerprint_server_url` now backfills its default URL rather than NULL). This is strictly more correct and matches what an ORM insert would write.
- CHANGELOG `[Unreleased]` entry added — this is a fix for shipped 0.26.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to Engram will be documented in this file.
 ### Fixed
 
 - **The manual import feature now works in Docker and other remote setups.** Browsing for files and starting an import over the network returned `403 Forbidden` because those endpoints were locked to the host machine. They now honor the **Allow LAN access** setting: when you've enabled it (headless/Docker installs enable it automatically on first run), import is reachable from another device, while the more sensitive fingerprint and contribution endpoints stay host-only. (#524)
+- **Upgrading to 0.26.0 no longer breaks the app with a 500 on every settings load.** On databases created before the customizable Discord messages feature, the new `discord_template_completed` / `discord_template_failed` columns were added without a default, leaving existing rows `NULL`. Because those fields are required text, `GET /api/config` then failed validation and returned HTTP 500 — which broke config loading across the whole dashboard (Settings, Contribute, first-run wizard). The schema reconciler now backfills new non-optional text columns with their declared default, and config responses tolerate legacy `NULL` values.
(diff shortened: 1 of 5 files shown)


Jsakkos:
@claude please review this PR

PR #559, opened 2026-07-29
fix(ci): retry the MakeMKV tarball download on transient Cloudflare errors

Jsakkos:
## Problem

[PR #554](https://github.com/Jsakkos/engram/pull/554) added retry logic to `docker/install-makemkv.sh`, but only to the **version lookup**:

```bash
curl -fsSL --max-time 30 --retry 3 --retry-delay 5 --retry-all-errors "${DL_BASE}/"
```

Its comment explains why `--retry-all-errors` is needed: makemkv.com sits behind Cloudflare, whose origin-unreachable codes (520-527) are not in curl's default retryable set.

The **tarball download** was left completely unprotected:

```bash
curl -fSL -o "makemkv-${part}.tar.gz" "${DL_BASE}/makemkv-${part}-${VERSION}.tar.gz"
```

That is backwards from a robustness standpoint. The tarball is the longer, heavier request against the same Cloudflare-fronted host, so it is *more* exposed to a transient 5xx, not less.

Observed on [PR #555](https://github.com/Jsakkos/engram/pull/555): `curl: (22) The requested URL returned error: 525` in the "Full MakeMKV install (1.18.3)" job.

## What changed

**1. Same retry treatment on the download**, plus a timeout sized for the payload:

```bash
curl -fSL --max-time 300 --retry 3 --retry-delay 5 --retry-all-errors \
    -o "makemkv-${part}.tar.gz" "${DL_BASE}/makemkv-${part}-${VERSION}.tar.gz"
```

`--max-time` applies **per attempt**, so 300s is a per-try budget for a multi-megabyte archive over a slow link, not a cap across all four attempts. The 30s used for the version lookup is right for a small HTML page and far too short here.

**2. A dedicated exit code**, following the existing `EXIT_FETCH_FAILED=2` / `EXIT_PARSE_FAILED=3` convention:

```bash
EXIT_DOWNLOAD_FAILED=4
```

Worth noting *why* this matters beyond tidiness: the old line had no error handling at all, so `set -euo pipefail` propagated **curl's own** status. The 525 in #555 surfaced as exit 22; a connection refusal surfaces as 7. That leaks curl's opaque numbering into the script's contract, which is exactly what those two constants exist to replace.

**3. Comments** mirroring the existing one, covering the Cloudflare rationale, the per-attempt timeout reasoning, and an explicit note that retries cover transient failures only.

## Scope check: other network calls

- `docker/entrypoint.sh` has **no** `curl`/`wget` calls at all. It only invokes this script, via `if ! ... ; then` (line 29), so it already tolerates any non-zero code.
- Those two `curl` calls in `install-makemkv.sh` are the only network fetches in `docker/`. Nothing else needed the same treatment.

## Please read: this will not fix a sustained outage

**Do not evaluate this by whether CI goes green.** During the incident that prompted it, the already-retry-hardened *version lookup* also failed, exiting after exhausting its three attempts, because makemkv.com's origin was down for an extended period rather than flickering.

The value is narrower and real: a brief hiccup mid-download currently fails the build immediately, and now it does not. The `MakeMKV install check` jobs will stay red until makemkv.com recovers. They are not required status checks.

## Verification

Verified by shellcheck and by exercising the failure path directly, not by waiting on CI.

**shellcheck 0.11.0** clean on both scripts, exit 0, zero warnings.

**Failure path exercised** by running the real script with `DL_BASE` repointed at a closed local port:

| | attempts | elapsed | exit | diagnostic |
|---|---|---|---|---|
| before | 1 | 2s | `7` (raw curl) | none |
| after | 4 (initial + 3 retries, 5s apart) | 23s | `4` | yes |

```
ERROR: could not download makemkv-oss-9.9.9.tar.gz after retries.
       Either makemkv.com is unreachable, or version 9.9.9 is no longer
       published (MakeMKV removes old releases). curl's error above
       distinguishes the two: a 5xx is an outage, a 404 is a stale version.
```

## One thing I deliberately did not change

The full-install path collapses a *version-resolution* failure to `exit 1` (lines 92-102), while codes 2/3 escape only through the `MAKEMKV_DETECT_ONLY` branch that `docker.yml` case-matches on. So after this change, "makemkv.com is down" reports as `1` from one stage and `4` from another, which is a little incoherent for a consumer asking "was this an outage?".

Making that uniform would mean changing the exit code of an existing path, which is outside what this PR is for, so I left it alone. Happy to do it as a follow-up if you want `makemkv-install.yml` to downgrade outages to warnings the way `docker.yml` already does.


Jsakkos:
@claude please review this PR

Issue #571, opened 2026-08-03
Add Delete/Reset Controls for Jobs and Allow Re-importing Previously Processed Folders


X-citizen-snips-X:
Engram is an awesome application, and its episode-identification workflow has already saved me a significant amount of time. However, the current job-management and re-import workflow is extremely frustrating and unnecessarily punishing when a user makes a simple mistake.

For example, I accidentally imported media as Season 1 instead of Season 3. After the incorrect job was created, I could not delete it, reset it, or retry the same folder. Engram continued reporting that a job for the folder already existed, even after the files had been moved or removed.

The only available workaround was to repeatedly rename the source folder so Engram would treat it as a new import. This resulted in:

Multiple unnecessary import jobs
Repeated folder renaming
Duplicate processing attempts
Additional disc ripping
Wasted time troubleshooting stale jobs
Engram complaining that files already existed after it had previously moved or removed them

A simple metadata mistake should not force the user to create increasingly ridiculous folder names or completely reset the application.

Requested Improvements

Please add:

A Delete Job button for failed, completed, archived, and incorrectly classified jobs
A Retry/Reprocess Job option
A Reset Import option that clears the stored folder-path association
A Re-import Anyway confirmation when a folder was previously processed
A way to correct the show, season, disc, or media type without creating a new job
Clear handling when destination files already exist
An option to keep, overwrite, rename, or skip existing files
A full Reset Database/History option in Settings

The current 409 Conflict behavior:

All import job(s) for this folder already exist (active or completed);
nothing new was started.

should present actionable choices instead of leaving the user trapped.

For example:

This folder was previously imported.

[Retry Existing Job]
[Delete Old Job and Re-import]
[Import as New Job]
[Cancel]
Why This Matters

Media ripping and episode matching naturally involve mistakes, ambiguous discs, duplicate playlists, incorrect season detection, and files that need to be re-ripped. Retrying and correcting jobs should be a normal part of the workflow, not something the application prevents.

Engram’s core matching functionality is excellent, but the lack of basic delete, retry, and reset controls makes normal recovery from mistakes much harder than it needs to be. This workflow needs guardrails and recovery options before it can be reliably used for processing large media collections.

X-citizen-snips-X:
Awesome, Thank you so much, I look forward to continue to provide Feedback on the tool.

PR #578, opened 2026-08-08
test: colocate useJobManagement test with its hook

Jsakkos:
## What changed

Moves `frontend/src/hooks/__tests__/useJobManagement.test.ts` to `frontend/src/app/hooks/__tests__/useJobManagement.test.ts`, next to the hook it tests.

- Three relative import paths updated (the only content change).
- Removed the emptied `src/hooks/__tests__/` directory so the fossil doesn't recreate the ambiguity.
- Updated the `TESTING.md` row, which pointed at the old path.

**No test logic changed.** This is a pure relocation.

## Why

The hook lives at `src/app/hooks/useJobManagement.ts`, but its test sat in a separate `src/hooks/__tests__/` tree. Three other hook tests are already colocated under `src/app/hooks/__tests__/`, so this file was the sole outlier across the repo's two parallel `hooks/` directories.

The split had a concrete cost: during work on #571, a search for existing coverage near the hook missed all 16 tests in this file, producing a false conclusion that `clearCompleted` was untested and nearly leading to a redundant test harness.

## Reviewer notes

The import fixes are asymmetric, which is the one thing worth a second look:

| Import | Before | After |
|---|---|---|
| hook under test | `../../app/hooks/useJobManagement` | `../useJobManagement` |
| `vi.mock` of `useWebSocket` | `../useWebSocket` | `../../../hooks/useWebSocket` |
| types | `../../types` | `../../../types` |

The hook import loses two levels (the subject is now adjacent) while the `vi.mock` path *gains* two, because `useWebSocket` genuinely still lives in `src/hooks/` and is now further away. This matters: `vi.mock` resolves relative to the test file and must match the module id the subject imports (`useJobManagement.ts` imports it as `../../hooks/useWebSocket`). A wrong mock path here fails quietly rather than loudly, so I verified the mock is still bound rather than trusting a green run: the tests make 21 references to the mock's captured `onOpen`/message callbacks, which would be `undefined` and throw if the real hook had loaded.

Three premise corrections found while doing this:

- `useDiscFilters` has **no** test at all, so it isn't the colocation precedent; the three other tests already in `src/app/hooks/__tests__/` are.
- The file has **16** tests, not 15. `TESTING.md` claimed 8; corrected to 16 since I measured it.
- No other mislocated tests exist. `src/hooks/__tests__/` held exactly one file, and the two hooks genuinely in `src/hooks/` (`useWebSocket`, `useSeasonRoster`) have no tests to move.

No CHANGELOG entry: internal test layout, no user-facing impact.

## Verification

| Check | Result |
|---|---|
| `npx vitest run src/app/hooks/__tests__/useJobManagement.test.ts` | 16 passed |
| Same file at old location, before the move | 16 passed (count unchanged) |
| `npm run build` | passed |
| `npm run test:unit` (full tier) | 45 files, 402 tests passed |
| `package-lock.json` | untouched (`node_modules` already present, no `npm install`) |

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -130,7 +130,7 @@ Run: `cd frontend && npm run test:unit` (~0.5s)
 | File | Tests | What It Covers |
 |------|------:|----------------|
 | `src/types/__tests__/adapters.test.ts` | 24 | Data transformation layer: `JobState` to UI state mapping (8 values), `TitleState` mapping (7 values), full job-to-disc-data transformation (TV/movie/fallback/unknown), duration formatting edge cases, match candidate extraction from JSON |
-| `src/hooks/__tests__/useJobManagement.test.ts` | 8 | WebSocket data merging logic: partial job update merging, title update targeting correct job/title, all-terminal-state detection, `titles_discovered` replacement, `subtitle_event` field updates |
+| `src/app/hooks/__tests__/useJobManagement.test.ts` | 16 | WebSocket data merging logic: partial job update merging, title update targeting correct job/title, all-terminal-state detection, `titles_discovered` replacement, `subtitle_event` field updates |
 
 ### Frontend E2E Tests (`frontend/e2e/`)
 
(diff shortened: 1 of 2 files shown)


Jsakkos:
@claude please review this PR

PR #579, opened 2026-08-08
fix: re-import a previously imported folder instead of blocking permanently (#571)

Jsakkos:
Fixes the import dead end reported in #571.

## The bug

Manual import refused any folder whose prior job was not `FAILED`. That rule was inherited from the polled watch folder deleted in #463: a poller fires unbidden and forever, so it needs a permanent memory of every path it has handled, but a human pressing Import has already stated intent. The comment above the guard still described the old intent and contradicted the code beneath it.

There was no escape. `DELETE /api/jobs/{id}` is a soft delete that leaves `staging_path` and `COMPLETED` intact, so the job vanished from the dashboard while still blocking re-import. `cancel_job` is a no-op on terminal states. `reset-all-jobs` is DEBUG only. Renaming the source folder really was the only remaining move.

A quieter second instance: `import_start` raised 409 only when *every* unit was skipped. A partial skip returned 200 with a `skipped` count no client read, so importing a show where seasons 1 and 2 were done and season 3 was new closed the modal having silently imported less than asked, with no error at all.

## What changed

**Path ownership replaces path history.** A staging path is *owned* by an in-flight job and merely *recorded* by a finished one. In-flight (including `REVIEW_NEEDED`, whose files still sit on disk) hard-blocks and is never overridable, because a second job would race the first. `COMPLETED` soft-blocks and is overridable by explicit per-unit confirmation. `FAILED` never blocks, as before. Rules live in the new `app/services/import_guard.py`.

This deliberately is not a copy of the disc-side guard, which lets `REVIEW_NEEDED` through: a disc in review has been ejected and the drive is free, whereas an import in review still holds its files.

**Per-unit conflict contract.** `POST /api/import/start` no longer returns 409 and no longer has a `skipped` count. It always returns 200 with `{job_ids, blocked[]}`, each blocked entry naming the unit, its reason (`in_flight` or `already_imported`), a display path, and the prior job ids. Requests accept `force_keys`, scoped to specific units so a confirmation about one unit can never force another. The key is an opaque hash, not a path, so content-derived identity can replace path-derived identity later with no client change.

**UI.** The import modal stays open on a conflict and renders an inline panel (not a nested modal, which in this app can leave a stuck overlay that swallows clicks). Soft-blocked units get Re-import Anyway; in-flight units are informational.

**Also fixed:** `POST /api/staging/import` was returning `{"status": "created", "job_id": <serialized object>}` when an import was actually refused. Failed jobs could not be dismissed from the dashboard at all: the clear action filtered to completed only, and its button was hidden when no completed jobs existed.

## Scope caveat, please read before closing #571

A library-mode import uses `shutil.move`, so a completed import leaves the source folder **empty**. Re-opening Import there yields `total_jobs: 0` and "No MKV files found here.", and the conflict panel is never reached. This change fixes the loop the reporter was actually stuck in (they had moved files back by hand, then hit the guard, then started renaming folders), but a user must still move files out of their library manually before re-importing. The CHANGELOG is worded to match.

Making `/api/import/preview` report "this folder was imported by job N, its files are in your library" would close that gap, but the spec commits to preview staying filesystem-only and DB-unaware, so it is tracked separately.

## Deliberately not included

Post-completion metadata correction (needs its own design: it means un-organizing files already in the library), full database/history reset, per-job delete buttons on cards, and per-file content identity. Requests #6 and #7 from the issue already ship as `conflict_resolution_default`.

## Test plan

- [x] Backend unit tier: 2661 passed, 19 skipped
- [x] Import-related backend suites: 48 passed
- [x] Frontend vitest: 417 passed across 46 files
- [x] `npm run build` and `npm run lint` clean
- [x] E2E `import-flow.spec.ts` green on two consecutive runs
- [ ] Manual: import a folder, let it finish, put the files back, re-import, confirm the panel offers Re-import Anyway and that pressing it starts a new job
- [ ] Manual: cancel a job to leave it failed, confirm CLEAR now dismisses it and it remains in History with its error

Every behavior change is mutation-tested: the guard's path scoping, the force scoping, the staging lock, `display_path`, the clear-button gate, and the stale-panel fix were each verified to turn a test red when reverted.


Jsakkos:
@claude please review this PR

PR #592, opened 2026-08-16
feat(discord): disc identity in notifications and review-needed alerts

Jsakkos:
## Summary

Two user-requested improvements to Discord notifications, plus the supporting work they needed.

**Disc identity.** Ripping a box set produced an identical notification for every disc, because the message text defaults to the show name, which is the same on disc 1 and disc 6. Notifications now carry structured Discord embed fields instead of one text blob: **Disc** (`THE_WIRE_S1 (Disc 3)`, falling back to a TheDiscDB slug, then a bare disc number), Season, Episodes, Duration, Tracks, Subtitles, Reason, Library. Empty values are dropped, so a movie shows no Season row. This works without anyone editing a template, including for users with a custom template saved before this change, because the template only supplies the description.

**Review-needed notifications.** A job parking in `REVIEW_NEEDED` was silent, so a walk-away rip stalled with nobody knowing. It now posts an amber "Review Needed" notification carrying the review reason. `REVIEW_NEEDED` is not terminal, so this hooks `on_transition` rather than `on_terminal_state`; that callback now receives `from_state` so a same-state re-broadcast does not re-ping, while a genuine re-park (review, re-identify, review again) does.

Also included: an episode manifest on completion (`S01E01-E04 (4 episodes)`, gaps preserved, ranges never spanning a season), an optional mention sent as message `content` (embeds do not resolve mentions, so this is what actually reaches a phone), an optional deep link to `/history/{job_id}`, per-event enable toggles, a "Send test message" button, and TMDB poster thumbnails.

## Notable details

- `notify_discord` swallows every error by default, which is load-bearing on the pipeline path. The test-webhook endpoint opts into `raise_on_error=True`, otherwise the button would report success for a revoked URL, the one thing it exists to rule out.
- Embed values are truncated to Discord's limits. A verbose MakeMKV stderr in `error_message` would otherwise 400 the request, and the swallow-all would silently cost the user the failure notification.
- `{{drive}}` keeps rendering the volume label (which it always did, despite the name) and is documented as a deprecated alias for the new `{{volume_label}}`. Changing its meaning would silently alter saved templates.
- `dashboard_base_url` is deliberately exempt from `is_safe_remote_url`: a LAN address is the expected value and the server never fetches it, only embeds it as a link. It has its own narrower guard against scheme injection and credential leakage.
- `POST /api/validate/discord-webhook` is gated to localhost/LAN. It is the only validator that POSTs to a caller-supplied URL rather than a fixed host.

## Test plan

- [x] Backend unit: 2811 passed, 19 skipped
- [x] Backend pipeline: 110 passed
- [x] Frontend unit: 428 passed across 46 files
- [x] `npm run build` (tsc + vite) clean, `npm run lint` clean, `ruff check` + `ruff format --check` clean
- [x] Single Alembic head (`b7d3f9a1c204`)
- [ ] Manual: configure a webhook, click **Send test message**, confirm a sample embed with a Disc field arrives
- [ ] Manual: simulate a TV disc through to COMPLETED, confirm the Disc field distinguishes disc number and the Episodes field lists what landed
- [ ] Manual: drive a job into REVIEW_NEEDED, confirm the amber notification pings the configured mention and links to the job


Jsakkos:
@claude please review this PR

Jsakkos:
Thanks. Addressed the review nit and, separately, the CodeQL failure.

**CodeQL `py/full-ssrf` (critical, alert #545)** at `discord_notifier.py` `client.post(webhook_url, ...)`. This was a genuine new taint path: before the test-webhook endpoint existed, `notify_discord` only ever received the stored config value, and now an HTTP request body reaches it. `is_safe_remote_url` did guard it, but a custom boolean guard is not something CodeQL can model as a sanitizer, and more to the point it only proved the host was *external*, not that it was *Discord*.

Fixed by following CodeQL's own recommended remedy rather than suppressing: `normalize_discord_webhook_url` parses a caller-supplied URL and rebuilds it from a hardcoded host plus an `int()` id and a character-class-restricted token, so the endpoint is structurally incapable of reaching any host but Discord. A non-Discord URL now returns a specific error instead of being posted to.

The stored-config path deliberately does not go through this: that value is gated by `is_safe_remote_url` on write, is not attacker-supplied at send time, and narrowing it would break anyone pointing Engram at a Discord-compatible relay.

**Item 1 (the `Library` field conditional)** parenthesized as suggested.

**Items 2 and 3** left as-is, agreeing with your read. On 2, dropping `Episodes` when nothing parses is intentional; the `Tracks` field still reports the count, so the completion isn't information-free. On 3, that is exactly the case the dedup rule cannot cover, and it is called out in `_notify_discord_on_review`'s docstring so a future caller updating `review_reason` in place has a chance of noticing.

Backend unit suite: 2818 passed, 19 skipped.

PR #601, opened 2026-08-19
fix: stop review jobs disappearing from every view

Jsakkos:
## What happened

A user reported that discs left parked in review for several days eventually vanished from the dashboard, and were not in job history either.

The rows were never deleted. I traced every path that can remove a `DiscJob`: `DELETE /api/jobs/{id}` and `DELETE /api/jobs/completed` are soft (they set `cleared_at`, and both reject non-terminal states), the only hard delete is `/api/simulate/reset-all-jobs` behind `require_debug`, there is no retention task (`cleanup_service` deletes staging directories, not rows), the startup stale sweep and the watchdog both exclude `REVIEW_NEEDED`, and the schema reconcilers use native `DROP COLUMN` without rebuilding tables.

What actually happened is that the job fell into a gap between two views:

- `GET /api/jobs` capped the dashboard at the 10 most recent uncleared jobs, ordered `created_at DESC`. Finished-but-uncleared jobs occupy slots too, so the window fills faster than it looks, and an old review job gets pushed off by newer discs.
- `GET /api/jobs/history` filtered to `COMPLETED`/`FAILED`, so a review job was excluded there by design.

Dashboard shows *recent and uncleared*; history shows *finished*. A job that is simultaneously old and unfinished is in neither set. `REVIEW_NEEDED` is precisely the state designed to sit still indefinitely, so it is the one state guaranteed to hit this.

## What changed

**Dashboard cap now applies to finished jobs only** (`list_jobs`). The 10 most recent terminal jobs are resolved first, then returned alongside every uncleared non-terminal job in a single `created_at DESC` sequence. Anything still waiting on the user or in flight stays on screen until it is dealt with; finished jobs still fall off after 10 and live on in history. Soft-delete still wins, so a cleared row stays hidden.

**History gained an explicit escape hatch** (`get_job_history`). The default is unchanged (completed + failed, which is what "history" means to a user), but an explicit `state` is now honoured for *any* `JobState`, and `include_all_states=true` drops the filter entirely. Together these make history a guaranteed backstop containing every job ever created.

**Frontend, so the backstop is actually reachable.** The History state filter gains **Needs review** and **All jobs**. This is slightly beyond the original scope, but fix 2 was inert without it, and making review jobs reachable exposed a second problem: the page rendered state as a binary in two places (`completed ? OK : FAIL`), which would have labelled a parked review job as "FAIL" and "Failed". Both now go through `historyState.ts`.

## Reviewer notes

- `state` on `/api/jobs/history` changed from `str | None` to `JobState | None`. This is a small behaviour change: a typo'd filter now 422s instead of silently falling back to the terminal-only default. The frontend only ever sent `completed`/`failed`, both still valid.
- `list_jobs` issues two queries instead of one. The first fetches only ids, and ordering stays in SQL so the caller still gets one authoritative sequence.
- Exempting non-terminal jobs cannot make the list unbounded in normal use: non-terminal jobs are bounded by the number of drives.
- Raising the cap instead would only have moved the cliff, not removed it. The cliff comes from mixing two populations under one recency rule: finished jobs grow without bound and belong in an archive, unfinished ones are bounded by hardware and must never scroll away.

## Testing

- `TestJobVisibilityInvariant` (9 cases) guards the invariant that every job is reachable from at least one surface. Four are guard-rails asserting behaviour that must *not* change: the cap still bounds terminal jobs, the newest 10 win the slots, default history stays terminal-only, cleared rows stay hidden.
- Red-green verified both directions: the 5 new tests failed before the fix, and `routes.py` was reverted to `HEAD` afterward to confirm they fail again before restoring. The failure message on the headline test is a literal reproduction of the report (`assert 1 in [13, 12, 11, ...]`, where job 1 is the review disc).
- 10 frontend cases across `historyQuery` and `historyState`.
- Backend: 2846 passed, 19 skipped. `ruff check` clean.
- Frontend: 450 passed across 48 files. ESLint clean, `tsc` + build succeed.
- **Not run locally:** the `tests/integration` tier, which executes against the real `engram.db` and organizes files into the real library. Two assertions there touch `GET /api/jobs`; both hold by inspection (a cancelled job is terminal and newest, so inside the 10-slot window; a soft-deleted job is excluded by the `cleared_at` predicate). CI covers them.

## For anyone already affected

Existing parked jobs are recoverable before taking this build:

```
sqlite3 ~/.engram/engram.db "SELECT id, volume_label, created_at FROM disc_jobs WHERE state='REVIEW_NEEDED' ORDER BY created_at;"
```

Each id opens at `/review/<id>` (`ReviewQueue` fetches the job directly, bypassing the capped list). After upgrading they reappear on the dashboard on their own.


Jsakkos:
@claude please review this PR

Jsakkos:
All three review findings were verified and fixed in 34835385.

1. **Duplicate terminal-states constant** — confirmed: `TERMINAL_JOB_STATES` is defined at `app/models/disc_job.py:23` and exported from `app.models`. Replaced `_TERMINAL_STATES` with it, materialized as `list(...)` to match the existing `arm_manual_identity` call site, and removed that function's now-redundant local import.
2. **Misleading timeline label** — confirmed: the row carries `completed_at`, which is null for a non-terminal job, so it rendered `Finished: -`. Now returns `In progress`; the test that locked in `Finished` was updated first and watched fail before the change.
3. **Stale REST docs** — confirmed, and the List Active Jobs section was stale too ("Limited to 10 most recent" no longer describes the behaviour). Both sections updated, plus an `include_all_states` row in the query-parameter table. `mkdocs build` passes with only the 18 pre-existing spec-link warnings.

Re-verified after the fixes and after rebasing onto v0.31.1: backend 2846 passed / 19 skipped, `ruff check` clean; frontend 450 passed across 48 files, ESLint clean, `tsc` + build succeed.

Also worth noting: the rebase surfaced the known CHANGELOG union-merge artifact. The driver appended this PR's `### Fixed` block inside the released `[0.31.1]` section rather than under `[Unreleased]`; that has been corrected.

PR #609, opened 2026-08-22
feat: manual review controls, combined-episode tracks, and structural disc detection

raiju:
Note: Heavily AI'd PR, made for personal use but figured it'd be useful for everyone. Really appreciate the work here!

## What prompted this

A six-disc box set of a segment-format cartoon ripped "successfully" and filed
**every episode of the season into `Extras/`**. Each disc reported COMPLETED, so
nothing surfaced the mistake and had to be fixed by hand.

The cause: TMDB numbers that show's ~7-minute segments as individual episodes,
while the DVD carries the assembled ~22-minute blocks. The duration pre-filter
compares each track against the season's episode runtimes and files anything that
doesn't match as bonus content. Judged one track at a time, it never notices that
it is about to reject *every* track on the disc.

Fixing that opened up a few adjacent problems, all around the same theme: **the
pipeline making confident decisions from weak signals, with no way for a human to
see or override them.**

## How would you like this handled?

This is larger than one feature, but don't know what your review preference is.
The history is already split into 13 logical/self-contained commits (lint-clean and
tested at the point it was made), grouped by
concern:

| | |
|---|---|
| `fix(matching)` | per-disc runtime trust policy |
| `fix(review)` | hold an all-extras disc |
| `feat(config)` | review every disc before organizing |
| `feat(episodes)` | one track holding several episodes |
| `feat(review)` | searchable episode picker + span |
| `feat(config)` | path checks + network shares |
| `feat(review)` | track preview |
| `feat(analyst)` | structural Play All detection |
| _(+5 smaller follow-ups)_ | |

**Happy to split this into separate PRs** along those lines — say four: the
extras/pre-filter fixes, manual review, combined episodes, and the
paths/preview work. Or squash it, or take it as-is. Tell me which you prefer and
I'll reshape it; I didn't want to open four PRs unasked.

---

### 1. Discs that fail silently

Two changes, independent of each other, both aimed at "looked successful, wasn't".

**The duration pre-filter now decides per disc how far to trust TMDB's runtimes**
rather than judging each track blind:

- *strict* — some track matches a single runtime. Today's rule, unchanged, so a
  disc holding real episodes **and** a long bonus feature still files that feature
  as an extra.
- *multiples* — no track matches a single runtime, but tracks match whole
  multiples of one. The show is segment-numbered; combined tracks count as
  episodes.
- *off* — nothing fits either way. The runtimes don't describe this disc, so skip
  the filter and let the audio matcher, which compares content rather than clocks,
  decide.

**An all-extras TV disc is held for review** regardless of that policy. "Nothing
on this disc is an episode" is far more often a failed match than a genuine bonus
disc, and confirming a real bonus disc costs one click where the alternative costs
a re-rip.

### 2. No way to take the wheel

A new setting holds **every** disc in the review queue for confirmation, however
confident the match was. The matcher still runs first. The check sits after the
escalation ladders, so review opens with its best guess pre-filled and you confirm
or correct rather than starting from nothing. Off by default.

### 3. One track really can be several episodes

Segment-format shows put three ~7-minute segments in a 22-minute slot, and TMDB
numbers each segment separately, so a single track genuinely *is* `S01E01-E03`.
Assigning it one code files it as `S01E01` and loses two episodes' worth of
identity.

`app/core/episode_codes.py` becomes the single place that reads an episode code. A
hyphen is a range, matching Plex/Jellyfin (and what hand-organized libraries mean
by it). Parsing is anchored on purpose: a local `S(\d+)E(\d+)` matched against a
combined code silently truncates it to its first episode, which is the bug class
this exists to prevent. Roster coverage counts a combined track against every
episode it claims, and such a track is deliberately **not** published to the
fingerprint network.

### 4. Assigning episodes was harder than it needed to be

The manual picker was a plain `<select>`, which is a bit annoying when a segment-numbered
season lists 100+ episodes. It now filters as you type over episode names and
numbers, using TMDB data already cached. A span control turns a pick into a range
for a combined track, pre-filled from the track's runtime against the season's.

Two smaller gaps in the same screen: the header named only the show and season
(which every disc in a box set shares, so reviewing disc 8 looked exactly like disc
7) and a track belonging to a **neighbouring season** couldn't be assigned at all,
since every code was built from the job's season. Both fixed.

### 5. You couldn't see what you were deciding about

Review asks "which episode is this?", and the fastest answer is to watch a few
seconds. Rips aren't browser-playable: DVD tracks are MPEG-2 with AC3 audio, so
repackaging doesn't help and the picture has to be re-encoded, so short clips are
transcoded on demand and streamed straight from ffmpeg's stdout. Nothing touches
disk; ffmpeg is killed when the client goes away. Seek points are the disc's own
chapter marks, which on a segment-format show land on the segment boundaries.

### 6. Configured paths failed silently too

Library and staging paths are the settings most likely to be quietly wrong — a
typo, an offline NAS, a container volume owned by another uid — and the mistake
surfaced as an organize failure *after* a completed rip. Each path field now
reports whether the folder exists, whether Engram can actually write to it (a real
touch/unlink probe, since mode bits on shares and container mounts routinely
disagree with reality), and how much space is free.

Network locations are treated as the normal case they are for a media library:
Windows UNC paths work wherever a local path does, `//server/share` and
`\\server\share` normalize to one stored value, an offline share is a warning
rather than a rejection, and a Windows-style share path on Linux/macOS explains
that it needs mounting. Note, this was only tested on Windows!

### 7. "Play All" detection was guesswork

A TV disc usually carries one long title playing every episode back to back;
ripping it wastes an hour and duplicates the disc. It was identified by summing
runtimes and looking for a title of roughly the right length — guesswork about a
question the disc can answer directly.

MakeMKV already reports how each title is assembled (`TINFO` attributes 25/26),
and Engram already parses and stores it. A Play All is a playlist over the episode
titles, so it reports several segments where each episode reports one:

```
idx=0  155.4min  chapters=37  segments=7  map=1-6,7-11,12-16,17-21,22-27,28-32,33-37
idx=1   22.3min  chapters=6   segments=1  map=1-6
idx=2   22.0min  chapters=5   segments=1  map=1-5
...
```

Those seven groups are `6,5,5,5,6,5,5` — exactly the chapter counts of titles 1–7,
in order. Proof of concatenation with no reference to runtime.

Because discarding a title wrongly costs a whole episode, **all three** signals
must agree: the segment count matches the map, the groups account for the
candidate's chapters *and* match a run of consecutive single-segment titles, and
the duration equals theirs combined within 2%. A two-part episode presented as one
playlist reports several segments too, which is why the chapter and duration
agreement is required rather than assumed. Discs reporting no segment data keep the
duration heuristic.

## Notes for review

- **Three migrations**, all additive defaulted booleans: `always_review`,
  `always_show_episode_span`, `play_all_detection_enabled`.
- **Four new settings**, all defaulting to current behaviour — the two review
  toggles and the span control are off, Play All detection is on (it replaces
  existing behaviour rather than adding it).
- **No breaking API changes.** `GET /api/jobs/{id}/season-roster` gains a
  `runtime` field per episode and now reports `season_count` for identified discs;
  new endpoints are `POST /api/validate/path` and the two preview routes.
- **ffmpeg is not bundled.** Previews use the same ffmpeg the chromaprint
  pre-decode already relies on, and degrade with an explicit message when absent.
- **One known limit:** an episode from a neighbouring season can't be matched
  automatically, since reference subtitles are only fetched for the disc's own
  season. It's a manual assignment by nature.


Jsakkos:
Thanks for this, and thanks for asking about the split up front rather than after the fact. The writeup is excellent and the commit history is clean enough that reshaping it is cheap. I read the whole diff and ran the new unit tests locally (203 passed).

**Short answer: yes please, split it.** Not because of the line count, but because two of these groups have defects I found that a focused review would have caught, and both are the "silently wrong forever" kind this PR is otherwise trying to eliminate.

## Proposed split and merge order

1. **All-extras hold + `always_review`** (`fix(review)`, `feat(config)`, the `always_review` migration). This alone fixes the bug you hit, it is small, and it is near-unfalsifiable. I would like to merge this first and quickly.
2. **Path checking and UNC normalization** (`feat(config)`, `paths.py`, `PathStatusHint`). Fully orthogonal, no pipeline coupling.
3. **Runtime trust policy** (`fix(matching)`).
4. **Multi-episode codes** (`episode_codes.py`, organizer, roster, contributions, picker and span UI).
5. **Structural Play All** on its own.

The preview commit comes out of the split entirely, see below.

## On the track preview

I have #610 in flight, which answers the same question by handing the file to a native player over a range-capable endpoint instead of transcoding. Rationale is in `docs/superpowers/specs/2026-08-22-review-external-player-design.md`, but the short version: ffmpeg is auto-detected and **not** bundled (`engram.spec` ships only `fpcalc`), so I do not want in-browser playback to depend on it, and I would rather not own an on-demand transcoder for what is an identification aid. VLC also exposes the disc's chapter marks itself, so the segment-boundary trick that motivated your chapter buttons survives the handoff.

Your version wins in two places I want to acknowledge, because they are real: reviewing from a phone or tablet on the LAN, and keeping watch-and-assign in one screen instead of an app switch per track. If it turns out people do want to triage the queue from a phone, in-browser preview is the right answer and I will revisit it, layered on #610's containment helper rather than starting fresh. For now please drop `feat(review): preview a track's video` from the split.

Practical note: #610 and your preview commit collide anyway. Both add routes under `/jobs/{job_id}/titles/{title_id}/...` and both edit `Inspector.tsx`, which yours rewrites.

## Findings

**Blocking for the Play All commit.** `analyst.py` short-circuits with `if any(t.segment_count for t in titles): return _structural_play_all(titles)`, so once any title reports segment data the duration fallback never runs. For Blu-ray the segment map is a clip list (`"1,2,3,4,5,6"`, per the existing comment at `extractor.py:1626`), which makes `groups` `[1,1,1,1,1,1]` and then requires the candidate to have exactly 6 chapters *and* six consecutive single-chapter episode titles. Real BD episodes have several chapters each, so I believe structural detection never fires there, and the heuristic that used to catch it is now skipped. `test_bluray_clip_list` covers `_parse_segment_groups` in isolation but nothing end to end. Could you show a real Blu-ray scan being detected before this lands? The DVD reasoning is convincing, it is the BD path I cannot verify.

**Blocking for the multi-episode commit.** `episode_codes.py` is not yet the single reader it claims to be: `discdb_exporter.py:36` still uses an unanchored `_EPISODE_RE.search`, so a combined track exports to TheDiscDB as `season=1, episode=1`. That publishes truncated metadata into a shared public database, which is worse than the internal case. `contribution_correction.py` and `disc_contribution_queue.py` were both handled deliberately and well, so this looks like a miss rather than a decision. Two smaller ones in the same group:

- `discord_notifier._EPISODE_CODE` is anchored `^S(\d+)E(\d+)$`, so a combined track counts toward the episode total but drops out of the range summary. Notifications will read like `S01E04 (4 episodes)`.
- In `organize_tv_episode`, the projection loop reassigns `out_season` each iteration, so a range projected across a season boundary silently takes the last episode's season for the whole filename. Projected numbers are also not guaranteed contiguous, which quietly turns a range into the run-on form.

**Paths.** `/api/validate/path` needs `require_localhost_or_lan`. The other validators in that file are gated when they have side effects or cost, and this one does a real touch/unlink at a caller-chosen path and reports existence, writability and free space. You also flagged that UNC handling is Windows-only tested, which is exactly where I would expect a cross-platform bug, so I would like a Linux pass on that PR.

## One design question

The three-mode runtime policy is the piece I am least sure about. `FILTER_MULTIPLES` is well argued and falsifiable. `FILTER_OFF` is the aggressive one, and now that the all-extras hold exists I am not sure it earns its keep: the hold already turns the catastrophic outcome into one click, without adding a per-disc mode that every future reader has to reason about. Two smaller notes if you keep it: the policy is cached on first call with no lock and is never invalidated on re-identify, where the runtime list changes underneath it; and `MAX_COMBINED_EPISODES_PER_TRACK=4` with a 25% over-tolerance accepts anything up to about 35 minutes against a 7 minute runtime.

Happy to be argued out of that one. Everything else here I want.

raiju:
Thank you for reading it that closely — the split is a better shape and I'll do it in the order
you listed. The preview commit is dropped; #610 is the right answer for the reason you give, and
I'd rather not own a transcoder either. If phone triage ever comes up I'll layer it on your
containment helper.

Two of your findings I went and checked against real scans rather than reasoning about. One ends
with me dropping the commit, the other with the fix being different from what either of us
assumed.

## Play All — you're right, and I'm dropping it

I sampled 386 disc scans out of `TheDiscDb/data` (every 6th `data/series/**/discNN.json`; 359 of
them have two or more episode items — 313 Blu-ray, 46 DVD) to check the BD path before arguing
for it. It doesn't survive the check, on two counts.

The detection is worse than you thought. It isn't only that structural never fires on BD — on
Blu-ray ordinary **episodes** routinely report several segments (30 Rock S1D1 has E3 as `"5,6"`,
E4 as `"7,8"`, E5 as `"9,10"`), so `any(t.segment_count for t in titles)` is true on essentially
every Blu-ray and that `return` at `analyst.py:975` disables the duration heuristic disc-wide.
Not a narrow regression.

The chapter-arithmetic test can't work there even in principle. A real BD Play All looks like
this:

```
idx=0  0:25:01  seg=2                          Episode S1E1
idx=1  3:19:44  seg=2,9,10,11,12,13,14,15      <- Play All
idx=2  0:24:58  seg=9                          Episode S1E2
...
idx=8  0:24:57  seg=15                         Episode S1E8
```

The map is exactly the set of clip ids of the eight episode titles — set containment, not chapter
counts. It also sits at index 1, between E1 and E2, so my contiguous-window search was wrong on
top of that. And containment can't be shared with DVD: MakeMKV reuses cell numbering across title
sets there, so on one sampled DVD two different 49-minute episodes both report `"1-4"`. Two
formats, two different algorithms.

Then I measured what a correct version would actually buy, and that's what settles it. Only 5% of
those 359 discs carry a Play All at all, and **15 of the 17 are 80 minutes or longer** — already
caught by the existing duration heuristic. Of the seven shorter ones the structural test would
add, five are *Courage the Cowardly Dog* tracks like "Serpent of Evil River / The Transplant" —
which aren't Play Alls at all, they're the 22-minute combined-episode tracks this PR exists to
keep. Deselecting those would be actively wrong.

So the honest accounting is: two format-specific algorithms plus a tuned duration floor, to gain
roughly two discs in 359, on a code path whose failure mode is losing an episode. Not worth it.
Commit dropped from the split.

I'd still like to leave the BD segment-map finding here in case it's useful to you later — the
clip-list form is a genuinely strong signal, it's just not one worth building a subsystem around
at this hit rate.

## TheDiscDB export — right, and the fix is the opposite of dropping the row

I started out thinking this one was already settled, since `_parse_episode_code`'s docstring ("For
multi-episode codes like 'S01E01E02', returns the first episode") predates the PR. That was wrong,
and the data repo is unambiguous about it. Scanning the same 386 files, 29 episode-bearing items
carry a non-numeric `Episode`, and the hyphenated range is the house convention for exactly this
case:

```
"17-18"  Courage the Cowardly Dog S2D2  "Serpent of Evil River / The Transplant"  0:22:43
"9-10"   The Expanse S1D2               "Critical Mass / Leviathan Wakes"         1:20:27
"7-8"    Supernatural S2D2              "The Usual Suspects / Crossroad Blues"    1:23:57
"1-2"    Season Six Disc One            "Deja Vu"                                 1:35:13
"5-7"    Downton Abbey Disc Two         "Play All"                                2:37:00
```

The Courage row is the segment-format cartoon that started this PR, recorded the way I'd want
Engram to record it. And the wire supports it end to end — `EngramEndpoints.cs` types
`season`/`episode` as `string?` behind `FlexibleStringConverter`, and it stays a string through
`UserContributionDiscItem` and `DiscItem` — so `"1-2"` serializes fine today.

So exporting `episode: 1` for a combined track isn't a safe under-count, it's a wrong value, and
dropping the row loses a layout the database wants. In the multi-episode PR I'll:

- have the exporter emit the range form for combined codes, via `episode_codes.parse_episode_code`
  so it stops being an unanchored `search` and the "single reader" claim is actually true;
- fix `disc_contribution_queue._derive_assignment`, which discards combined tracks as
  `"discarded"` — same reasoning, opposite direction, and it drops a genuine Episode row;
- update `tests/pipeline/test_export_schema.py`, which currently asserts `episode` is an `int`.

Worth flagging since it changes the published shape: single-episode tracks stay integers, only
combined ones become `"1-2"` strings. Say the word if you'd rather they were strings uniformly.

## The rest

`organize_tv_episode` and `discord_notifier` are both straightforward misses, no argument —
`out_season` reassigned inside the projection loop, and `^S(\d+)E(\d+)$` matched against codes that
are now ranges. Both get fixed in the multi-episode PR with tests that would have failed.

`/api/validate/path` gets `require_localhost_or_lan`; you're right that it's the one endpoint in
that file doing real filesystem work at a caller-chosen location. On the Linux pass — I found one
before running it. `is_unc_path` treats any leading `//` as a Windows UNC, but `//mnt/media` is a
legal POSIX absolute path, so on Linux it gets rejected with "mount the share and use its mount
point instead" for a path that works. Fixed and covered in that PR, along with a real Linux run
rather than my asserting it's fine.

## FILTER_OFF — dropping that one too

You're right, and once I went through it case by case it's clearer than "the hold makes it
redundant".

The trigger overlap is total, as you say: FILTER_OFF fires only when no track fits under either
rule, which means every track routes to `_handle_extras`, which means the all-extras hold fires.
So it was never the thing standing between the user and the bad outcome — at most it decides
whether the review opens with the matcher's guesses pre-filled or blank.

And for most of the reasons the runtimes can be wrong, the matcher has nothing to offer either.
If we matched the wrong same-name show, or the wrong season, ASR is working from that same wrong
identification and fails too — review either way, pre-filled with nothing. The only case where
skipping the pre-filter genuinely rescues a disc is the one where the show and season are right
and TMDB's runtime figures aren't.

Which turns out to have a much smaller fix. `fetch_season_episode_runtimes` builds the list as
`[ep.get("runtime", 0) or 0 for ep in episodes]`, so a season TMDB carries no runtimes for comes
back as `[0, 0, 0, …]` — truthy, so the `if runtimes` guard at `matching_coordinator.py:839`
passes, `_matches_expected_runtime` filters the zeros away to nothing, and every single track
fails the check. That is a genuine all-extras path, and `if any(r > 0 for r in runtimes)` closes
it permanently without a third per-disc mode for future readers to reason about.

So: FILTER_OFF comes out, the zero-runtime guard goes in, FILTER_MULTIPLES stays — it's the one
that's well argued and falsifiable, and it's the mode the disc that started this actually needed.
The hold stays the floor underneath both.

Your three smaller notes, with FILTER_OFF gone:

- **Staleness is real**, and it predates the change — nothing clears `_episode_runtimes` either
  outside terminal states, so a re-identify keeps the previous show's runtimes regardless of the
  policy sitting on top of them. I'll clear both on re-identify in that PR.
- **The lock I'd argue against**: the policy is a pure function of the same runtimes and the same
  selected durations, so a concurrent double-compute reaches the same verdict. Now that it's a
  two-way decision rather than three, I'd rather say that in a comment than take a lock.
- **The 35-minute window is the one that deserves the fix.** `4 × 7min × 1.25` accepts anything up
  to 35 minutes against a 7-minute runtime, which is far too wide to call corroboration. I'll bound
  the combined window in absolute terms rather than as a percentage that grows with `n`.

That leaves four PRs rather than five: all-extras + `always_review`, paths, runtime policy,
multi-episode. I'll start on the first now since it's unchanged and you want it quickly. One
mechanical note for scheduling: the remaining two Alembic migrations land in PRs 1 and 4, so the
second needs its `down_revision` re-parented once the first merges — they'll want to go up in
sequence rather than in parallel.

raiju:
Split done — closing this in favour of the four PRs below, in your merge order.

| | PR | State |
|---|---|---|
| 1 | #611 — all-extras hold + `always_review` | **Ready for review** |
| 2 | #612 — path checking and UNC normalization | Draft |
| 3 | #613 — runtime trust policy | Draft |
| 4 | #614 — multi-episode codes | Draft |

**#611 is the one to look at.** It is unchanged from the commits you already read and ran, and it
has had a human pass on this end — that is the one I would merge first, as you suggested.

**The other three are drafts on purpose.** They carry the review fixes, but the changes I made in
response to your findings have not had a human read yet on my side — only the test suites. Please
treat them as work-in-progress until I mark them ready. I did not want to present machine-written
follow-ups to your findings as though they had been checked properly.

Dropped from the split entirely, as discussed above: the track preview (yours in #610 is the right
answer) and structural Play All detection. On the latter — I measured a correct version against 359
real disc scans from `TheDiscDb/data` before arguing for it, and it does not earn its place: only
5% of discs carry a Play All at all, 15 of the 17 are already caught by the existing duration
heuristic, and 5 of the 7 it would add are *Courage the Cowardly Dog* combined-episode tracks it
would be actively wrong to deselect. `FILTER_OFF` went the same way for the reason you gave, with
the one case it really did rescue fixed properly instead.

Each PR describes what changed since your review. In short: #612 gains the localhost/LAN guard plus
a POSIX `//path` fix I found before running the Linux pass; #613 loses `FILTER_OFF`, gains a
usable-runtime guard, a bounded combined window and cache invalidation on re-identify; #614 fixes
the exporter (range form, not truncation — TheDiscDB's own convention), the contribution queue, the
Discord manifest and the cross-season projection.

#614 is stacked on #611 and carries its commits until that one lands; I will rebase it then.

Thanks again for the review!

PR #611, opened 2026-08-23
fix(review): hold an all-extras disc, and a setting to review every disc

raiju:
First of the four PRs split out of #609, per the review there. This is the one you asked to
merge first and quickly — it is unchanged from the commits you already read.

## What it does

Two independent changes, both aimed at "looked successful, wasn't".

**An all-extras TV disc is held for review.** A disc where nothing matched an episode and
everything was auto-sorted into `Extras/` is a matching failure wearing a success costume: it
organizes cleanly, reports COMPLETED, and the mistake stays invisible until someone browses the
library weeks later. Across a box set that is every episode of the season. "Nothing on this disc
is an episode" is far more often a failed match than a genuine bonus disc, so the disc now parks
in review. Confirming a real bonus disc costs one click; the alternative costs a re-rip.

This is the floor, not an override — it fires independently of the setting below.

**A new `always_review` setting** (Settings → *Naming & extras*) holds every disc in the Review
Queue for confirmation, however confident the match was. The matcher still runs first, so the
review opens with its best guess pre-filled; you just get the last word before anything moves
into the library. Useful for shows the matcher handles badly.

It is checked *after* the escalation ladders, so the review is pre-filled rather than blank, and
gated on `has_matched` — a disc with nothing left to organize must not be parked, or the review
page has no way to finish it.

## Notes

- Migration `c3e17a2b8d40_add_always_review_override` adds the one config column. It is the first
  of the two migrations in the split; #609's other migration (episode span) is parented on this
  one, so this PR wants to land before that one.
- The gitignore commit rides along: it globs the backend database ignore rules so
  `engram-b.db` and friends from a second worktree stop showing up as untracked.

## Tests

`tests/unit/test_all_extras_review.py` and `tests/unit/test_always_review.py`, both new. The
all-extras tests fail against `main`.


Jsakkos:
@claude review this PR

PR #612, opened 2026-08-23
feat(config): check configured paths and support network shares

raiju:
Second of the four PRs split out of #609. Fully orthogonal to the others — no pipeline coupling,
no migration.

## What it does

Settings paths are the ones most likely to be quietly wrong: a typo, a share that isn't mounted, a
Docker volume owned by another uid. Left unchecked the mistake surfaces hours later as an organize
failure after a completed rip, which is the worst possible moment to learn about it.

`app/core/paths.py` answers "will Engram actually be able to use this path?" — exists, is a
directory, is writable, how much room is left, and is it a network location. The write check is a
real touch/unlink probe rather than a permissions calculation, because on network shares and
container mounts the mode bits routinely disagree with what the server can actually do. It never
raises: a settings screen asking "is this path ok?" must always get an answer, even for a nonsense
value.

Network locations get first-class treatment because they are the normal case for a media library.
On Windows that means UNC paths (`\\server\share`), normalized so `//server/share` and
`\\server\share` are stored as one setting rather than two that look different and behave the
same. An unreachable share is reported as *unreachable* rather than *invalid* — saving it is
reasonable, using it today is not.

`PathStatusHint` surfaces the result inline under each path field in the Config Wizard.

- **`POST /api/validate/path` now requires `require_localhost_or_lan`.** You were right that it is
  the one endpoint in `validation.py` doing real filesystem work at a caller-chosen location — it
  touches and unlinks a probe file and reports existence, writability and free space. It is now
  gated like the other two.
- **Fixed a POSIX bug in `is_unc_path` before running the Linux pass.** It treated any leading
  `//` as a Windows UNC path, but `//mnt/media` is a legal POSIX absolute path, so on Linux a
  perfectly good path was rejected with "mount the share and use its mount point instead". UNC
  detection is now Windows-only, and `//path` on Linux is treated as the ordinary absolute path it
  is.
- Ran the suite on Linux rather than asserting it was fine.

## Tests

`tests/unit/test_paths.py` (normalization, UNC handling on both platforms, the unreachable-share
case, probe cleanup), the new route tests in `tests/unit/test_api_routes.py` including the
localhost/LAN gate, and `PathStatusHint.test.tsx` on the frontend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,49 @@ All notable changes to Engram will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Review every disc before it is organized.** A new setting under *Naming &
+  extras* holds every disc in the Review Queue for confirmation, however
+  confident the match was. The matcher still runs and pre-fills its best guess —
+  you just get the last word before anything moves into the library. Useful for
+  shows the matcher handles badly.
+- **One track can now be assigned several episodes.** Segment-format shows
+  (three ~7 minute cartoon segments in one 22 minute DVD track) list each segment
+  as its own TMDB episode, so a single track really is `S01E01-E03`. The review
+  picker can now say so, and the file is named with the range form Plex and
+  Jellyfin both read.
+- **The manual episode picker is searchable.** Type part of an episode name or
+  its number instead of scrolling a dropdown — which matters for the 40-100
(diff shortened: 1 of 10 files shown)


Jsakkos:
@claude review this PR

PR #613, opened 2026-08-23
fix(matching): decide per disc how far to trust TMDB runtimes

raiju:
Third of the four PRs split out of #609, reshaped per the review there — `FILTER_OFF` is gone.

## The problem

Before Engram listens to a track, it checks the track's length against TMDB's episode runtimes for
the season. A track whose length matches nothing is called bonus content and filed into `Extras/`
without ever being matched. That check exists so a 6-minute featurette doesn't get forced onto an
episode slot, and for an ordinary disc it does its job.

It fails badly when TMDB's runtimes describe something other than the disc's tracks. Judged one
track at a time, the rule cannot see that it is about to reject *every* track on the disc — so a
segment-format cartoon, where TMDB numbers the ~7-minute segments and the DVD carries assembled
~22-minute blocks, files its entire season into `Extras/` and reports success.

## What this does

`MatchingCoordinator._runtime_filter_policy` decides **once per disc**, from the whole disc rather
than one track:

- **strict** — some track matches a single runtime. Today's rule, unchanged, so a disc holding
  real episodes *and* a long bonus feature still files that feature as an extra.
- **multiples** — no track matches a single runtime, but tracks match whole multiples of one. The
  show is segment-numbered; combined tracks count as episodes and the span gets assigned in review.

The strict rule stays in force whenever *any* track matches a single runtime, so this is a per-disc
escape rather than a general widening of the window.

- **`FILTER_OFF` is dropped.** You were right that the trigger overlaps the all-extras hold
  completely, and going through it case by case it's worse than redundant: for the wrong-show and
  wrong-season cases, ASR is working from that same wrong identification and fails too, so the
  review opens blank either way. It bought a third per-disc mode for one narrow case.
- **That one case gets a real fix instead.** `fetch_season_episode_runtimes` builds the list as
  `[ep.get("runtime", 0) or 0 for ep in episodes]`, so a season TMDB carries no runtimes for comes
  back as `[0, 0, 0, …]` — truthy, so the `if runtimes` guard passed, `_matches_expected_runtime`
  filtered the zeros away to nothing, and every track failed. The guard now requires at least one
  usable runtime, which closes that all-extras path permanently and without a mode.
- **The combined window is bounded absolutely.** `4 × 7min × 1.25` accepted anything up to 35
  minutes against a 7-minute runtime, which is too wide to call corroboration. The over-tolerance
  no longer grows with the number of parts.
- **Caches are cleared on re-identify.** Neither `_episode_runtimes` nor the policy derived from it
  was invalidated when a re-identify changed the show underneath them — pre-existing for the
  runtimes, but the policy made it visible. Both are cleared now.
- No lock on the policy cache: it is a pure function of the same runtimes and the same selected
  durations, so a concurrent double-compute reaches the same verdict. Said in a comment rather than
  paid for with a lock.

## Tests

`tests/unit/test_matching_coordinator.py` — the policy decision on each disc shape, the
zero-runtime guard, the bounded combined window, and cache invalidation on re-identify.


Jsakkos:
@claude review this PR

PR #614, opened 2026-08-23
feat(episodes): support one track holding several episodes

raiju:
Fourth of the four PRs split out of #609.

**Stacked on #611.** Its migration is parented on #611's, so this branch carries #611's three
commits as well — GitHub can't target a base branch that lives in the fork. Merge #611 first and
I'll rebase; this diff then shrinks to its own six commits plus the review follow-ups. Review the
last seven commits, or use the compare against `split/review-hold` if that's easier.

## What it does

One track can hold several episodes. This is real for segment-format shows, where TMDB numbers the
~7-minute segments and the DVD carries the assembled ~22-minute block, and for ordinary two-part
episodes shipped as a single title.

`app/core/episode_codes.py` becomes the one place that parses and canonicalizes episode codes, so a
combined code like `S01E01-E03` survives everything downstream instead of being silently truncated
to its first episode by a local `S(\d+)E(\d+)`. `organize_tv_episode` names the file with the
Plex/Jellyfin range form, and roster coverage counts the track against every episode it claims.

The review UI gets what's needed to drive it: a searchable episode picker, a span control for
assigning a range to one track, a setting to always show that control, the disc's name in the
review header, and the ability to assign a track to a neighbouring season.

- **The DiscDB export now emits the range form.** You were right, and after checking `TheDiscDb/data`
  the fix turned out to be the opposite of dropping the row — the hyphenated range is the house
  convention for exactly this case (`"17-18"` on *Courage the Cowardly Dog*'s
  "Serpent of Evil River / The Transplant", `"9-10"` on *The Expanse*, `"7-8"` on *Supernatural*,
  `"1-2"` on a two-parter). `_parse_episode_code` now goes through `episode_codes.parse_episode_code`
  rather than an unanchored `search`, so the single-reader claim is actually true.
- **`disc_contribution_queue._derive_assignment` no longer discards combined tracks.** Same
  reasoning in the other direction: typing a genuine Episode row as `"discarded"` drops a layout the
  database wants.
- **`discord_notifier._EPISODE_CODE`** was anchored `^S(\d+)E(\d+)$` and matched against codes that
  are now ranges, so a combined track counted toward the episode total but dropped out of the range
  summary — notifications read `S01E04 (4 episodes)`. It goes through `episode_codes` too.
- **`organize_tv_episode`'s projection loop** reassigned `out_season` on each iteration, so a range
  projected across a season boundary took the last part's season for the whole filename. The
  projection now requires every part to land in one season and falls back to canonical numbering
  when it doesn't, rather than writing a filename that quietly claims the wrong season.
- `tests/pipeline/test_export_schema.py` updated: single-episode tracks stay integers, combined
  ones are the `"1-2"` string form. The wire accepts both — `EngramEndpoints.cs` types
  `season`/`episode` as `string?` behind `FlexibleStringConverter`.

## Found by sweeping the rest

Your DiscDB finding prompted me to check every reader of `matched_episode`, and every field typed
as a single episode number, rather than just the sites you named. Four more, all the same mistake:

- **The acoustic fingerprint queue** parsed the code with a prefix-anchored regex and contributed
  the first episode. A combined track's audio is several episodes and a fingerprint row names one,
  so this published a wrong identity into the shared network — worse than the DiscDB case, since
  every later ripper of that disc would inherit it. There is no honest single-episode label for
  that audio, so those tracks now contribute nothing.
- **`_same_episode_code`** compared first episodes only, so the "already organized by another
  track" warning missed everything after a combined track's first: assigning `S01E02` next to an
  existing `S01E01-E03` raised nothing. It tests set overlap now.
- **TheDiscDB ingest**, symmetric to the export. `int("17-18")` raises, so those mappings arrived
  with no episode at all — the database knew the answer for precisely the tracks that are hardest
  to guess, and we discarded it. `DiscDbTitleMapping` carries the whole claim now; the existing
  `episode` field stays the first number so mappings persisted before this still deserialize.
- **The history amend picker** built its episode list with `E(\d+)$`, taking only the *last*
  episode of a combined code, so earlier ones vanished from the choices.

One deliberate non-fix: `bootstrap_library.parse_episode_filename` does not parse a combined
filename, so `Show - S01E01-E03.mkv` is reported unparseable rather than bootstrapped. That is the
right outcome for the same reason as the fingerprint queue, and it is now commented and tested so
nobody "fixes" it by loosening the pattern.

## Note on the published shape

Only combined tracks change: a single-episode track still exports `episode: 1`. Say the word if
you'd rather they were strings uniformly and I'll make it consistent.

## Tests

`tests/unit/test_episode_codes.py`, `test_naming_conventions.py`, `test_contribution_correction.py`,
the export schema pipeline tests, and on the frontend `EpisodePicker.test.tsx`,
`Inspector.test.tsx`, `coverage.test.ts`, `utils.test.ts`.


Jsakkos:
@claude review this PR

Jsakkos:
Thanks for this, @raiju. The episode-code work here is the right design, and we're going to build on it.

A user just hit this exact case with a Dexter's Laboratory set: three ~7-minute segments per 22-minute track. Your approach is what we want for it: store the combined code (`S01E01-E03`), read it everywhere through one anchored parser in `episode_codes.py`, and name the file in the Plex/Jellyfin range form. Your sweep of every place that reads episode codes held up when I checked it against current main. It covers the fingerprint skip, TheDiscDB's `"17-18"` form in both directions, overlap-based duplicate checks, and falling back to aired numbering when DVD ordering splits a range across seasons.

We won't merge this PR as-is, for three reasons:

- **It's out of date.** It conflicts with main in 7 files and still carries #611's commits.
- **It predates #624.** That PR detects combined tracks from where the audio matches land. It isn't wired in here, so a confirmed detection never becomes a combined assignment, and the "cannot name a combined file yet" message stays.
- **The earlier bot review's collision finding is still open.** `TitleList.tsx` and `ReviewQueue.tsx` check `collisions.has(selection)` against single-episode codes, so a combined track that overlaps another track is never flagged.

The plan: combined naming goes in as a separate PR built on `episode_codes.py`, the organizer range splice, and your reader fixes, credited to you in the changelog. That PR also:
- pre-fills detected combined tracks for confirmation;
- fixes the collision check;
- rebases onto current main.

I'm leaving out the review UI additions (searchable picker, span control and its setting, disc name in the review header, neighbouring-season assignment) to keep that PR reviewable. Any of them would be welcome as smaller PRs once combined naming lands. I'll link the PR here when it's up.

PR #635, opened 2026-09-04
fix(subtitle-cache): a non-ASCII show title no longer aborts a harvest

Jsakkos:
## Summary

A 448-show cache build died at **show 179 of 448** on `Shōgun` (TMDB 126308). Windows opens both a legacy console and a redirected pipe with the ANSI codepage (cp1252), which cannot encode `ō`. Rich repaints the progress bar through that stream, so a single title raised `UnicodeEncodeError` and killed the run:

```
File "rich/_win32_console.py", line 402, in write_text
File "encodings/cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '\u014d'
```

**The harvest itself was fine. Only the rendering of its progress was fatal.** 178 shows had already been harvested successfully; the crash landed before packaging, so the tarball was never written and all of that work went unpublished. The coverage rows and downloaded SRTs survived on disk, so nothing was lost permanently, but the run had to be redone.

`main()` now forces stdout and stderr to UTF-8 with `errors="replace"` before anything renders. An unrepresentable character degrades to a placeholder instead of ending a multi-hour run.

Three other titles in `curated_shows.csv` carry the same hazard (`El Señor de los Cielos`, `Sin senos sí hay paraíso`, `Pasión de Gavilanes`), so this was not a one-off.

## Why `errors="replace"`

Switching to UTF-8 alone would fix the titles we know about. `errors="replace"` is the half that makes it permanently non-fatal: some future stream that cannot represent a character produces a mangled glyph rather than ending the run. For a script that is about to run unattended on a nightly timer, a cosmetic defect is strictly better than a dead build.

Streams that cannot be reconfigured (already detached, or a test double without the method) are skipped rather than raised on. Failing to change an encoding must never be worse than the crash it prevents.

## Verification

Reproduced and fixed against the real failure, not a synthetic one:

```
uv run python scripts/build_subtitle_cache.py --show-list shogun.csv ... > log 2>&1
EXIT=0
Built cache: 1 shows, 10 episodes, 0.1 MB
  from disk:         10 (1 covered seasons shipped without network)
  OS quota left:     1000 downloads today
```

Exit 0, tarball built, and as a bonus this confirms the complete-on-disk fast path from #631: `Shōgun`'s SRTs were harvested moments before the original crash, and `is_done` shipped all 10 episodes without spending a single download.

- [x] 5 new unit tests covering the cp1252 stream, the `replace` handler, a stream with no `reconfigure`, a stream that refuses it, and the multi-stream call
- [x] `tests/unit/test_build_subtitle_cache.py` and the three sibling suites: **83 passed**
- [x] `ruff check` / `ruff format --check` clean
- [x] Live reproduction confirms the fix

## Note

This is not reachable on Linux, so the eventual Ubuntu server would not have hit it. It blocks harvesting from the Windows laptop, which is where the corpus lives today.

### DIFF
--- a/backend/scripts/build_subtitle_cache.py
+++ b/backend/scripts/build_subtitle_cache.py
@@ -143,6 +143,36 @@ class RunTally:
         return self.cache_hits / denom if denom else 0.0
 
 
+def _ensure_utf8_output(*streams) -> None:
+    """Force the given text streams to UTF-8 so a show title cannot abort a run.
+
+    On Windows both a legacy console and a redirected pipe default to the ANSI
+    codepage (cp1252), which cannot encode titles like "Shogun" (spelled with a
+    macron) or "El Senor de los Cielos". Rich repaints the progress bar through
+    that stream, so ONE such title raised UnicodeEncodeError and killed a real
+    448-show harvest at show 179, discarding the packaging step for the 178
+    shows that had already succeeded. The harvest itself was fine; only the
+    rendering of its progress was fatal.
+
+    ``errors="replace"`` is the half that makes this permanently non-fatal: a
+    character some future stream cannot represent degrades to a placeholder
+    rather than ending a multi-hour run.
+
+    Streams with no ``reconfigure`` (a test double, a plain object) are skipped,
+    and a stream that refuses to be reconfigured (already detached) is left
+    alone. Failing to change an encoding must never be worse than the crash it
+    is preventing.
+    """
+    for stream in streams:
+        reconfigure = getattr(stream, "reconfigure", None)
+        if reconfigure is None:
+            continue
+        try:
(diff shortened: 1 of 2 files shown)


Jsakkos:
@claude please review this PR

Jsakkos:
Acted on the non-blocking nit from the second review: the `except` is now broad, because the docstring promised that failing to change an encoding is never worse than the crash it prevents and a `TypeError` from a malformed `reconfigure` attribute would have broken that promise. It runs once at startup, so a broad catch costs nothing and cannot mask a later error. Added two tests for the shapes the narrow catch missed (non-callable `reconfigure`, and one raising an arbitrary exception). 36 tests pass in that file.

PR #636, opened 2026-09-04
fix(subtitle-cache): treat a scraper outage as unmeasurable, not as zero coverage

Jsakkos:
## What

Closes the scraper half of the harvester's degraded detection, the gap recorded as
deliberately out of scope in `docs/superpowers/plans/2026-08-31-subtitle-cache-harvester-repair.md`.

PR #631 taught the cache builder to decline to record a coverage row when a season could
not actually be measured: **no record beats a false zero**. But `_is_degraded` read only
the OpenSubtitles half of the cascade, and said so in its own docstring. With
OpenSubtitles healthy and BOTH Addic7ed and TVsubtitles down, every episode came back
`not_found`, `os_failed` was False, the season recorded at 0%, and `should_skip`
suppressed it for 30 days. Same defect, different door.

## How

**The scheduler already knew.** `_CircuitBreaker.record_failure` counts only exceptions;
a clean miss or an unusable download calls `record_success` because the provider
*responded*. That is exactly "had nothing" vs "never answered", and it was being thrown
away. `run_jobs` now returns a `SchedulerRun` (`.results` + `.provider_health`) instead of
a bare dict, and `_is_degraded` gains a `providers_failed` axis beside the OpenSubtitles
one.

Design decisions, with reasoning rather than guesses:

- **Keyed to the breaker trip, not to `transport_failures > 0`.** A lone scraper exception
  is routine. Flagging on it would degrade nearly every season, stop coverage being
  recorded at all, and walk straight into the other known gap ("a persistently degraded
  corpus is re-measured forever"). Three consecutive transport failures is the
  already-tuned "this provider is down" verdict.
- **Partial results follow PR #631's asymmetry, which I checked rather than assumed.**
  One dead scraper plus one healthy one that covered the whole season records normally:
  there is no gap left for the outage to have hidden. Incomplete after a provider went
  dark writes no row, because a partial season cannot tell "genuinely absent" from "was
  behind the dead provider". Over-flagging costs one cheap retry; under-flagging costs a
  30-day blind spot.
- **Breaker-open for a season never attempted is unreachable today**, because `run_jobs`
  builds a fresh `Scheduler` per call and `download_subtitles` calls it once per season.
  A test pins that. Making breakers process-global is a genuinely independent follow-up,
  but it would make the question live, so the reasoning is left in a comment at the
  decision point.
- **`SchedulerRun` is a dataclass, not a `dict` subclass carrying an attribute.** The
  subclass needs zero test churn, but a mocked plain `{}` would then silently report "no
  providers failed" through a `getattr` fallback, reintroducing the exact silent-underflag
  mode this closes. Breaking the seven stale mocks loudly is the point.
- **The precomputed-gap heal is included.** It copies `degraded: False` forward from the
  skip dict, so a heal attempted while the scrapers were down used to report a confident
  (and short) season. `_fetch_episodes` returns its provider health and
  `_heal_precomputed_gaps` applies the same rule.

## Testing

`_is_degraded` truth table over the new axis, `run_jobs` health against dead / missing /
transiently-failing / mixed clients, `download_subtitles` end to end for the outage and
its control, the heal path, and a build-script-level test that runs the real
`download_subtitles` under `_harvest_show` and asserts **no coverage row is written** for
a scraper outage while a healthy-but-empty run still writes its 0% row.

The three defect-catching tests were verified to fail against the pre-fix decision while
the six controls still pass, so they are not asserting `True`. No real cache build was
run; no OpenSubtitles quota consumed.

`uv run pytest tests/unit`: **3199 passed, 21 skipped**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ All notable changes to Engram will be documented in this file.
   that exhausted the quota in the first place. Run summaries also credit
   OpenSubtitles correctly instead of reporting its downloads as cache hits.
 
+- **The same protection now covers the subtitle scrapers.** The fix above read
+  only OpenSubtitles' health, so a build where OpenSubtitles was fine but both
+  Addic7ed and TVsubtitles were unreachable still wrote those seasons down as
+  having no subtitles and stopped retrying them for 30 days. A scraper that
+  goes down is now told apart from one that simply has nothing, so a season
+  nobody could reach is left unmeasured and picked up on the next run, and the
+  run summary names the provider that never answered.
+
 ## [0.34.0] - 2026-08-31
 
 _Highlights: run identification and episode matching entirely on your own machine with Ollama or LM Studio._
--- a/backend/app/matcher/provider_scheduler.py
+++ b/backend/app/matcher/provider_scheduler.py
@@ -16,7 +16,11 @@ Public surface:
 - ``EpisodeJob`` — dataclass describing one episode's lifecycle.
 - ``ProviderWorker`` — thread that pulls jobs for one provider.
 - ``Scheduler`` — wires workers together and waits for completion.
-- ``run_jobs(jobs, workers)`` — convenience entry point.
+- ``ProviderHealth`` / ``SchedulerRun`` -- per-provider reachability, so a
+  caller can tell "this provider had nothing" from "this provider never
+  answered".
+- ``run_jobs(jobs, workers)`` — convenience entry point; returns a
+  ``SchedulerRun``, not a bare results dict.
 
 Worker contract (any provider):
 - ``client.get_best_subtitle(show, season, episode) -> entry | None``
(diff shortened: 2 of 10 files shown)


Jsakkos:
@claude please review this PR

Jsakkos:
Both reviews addressed. One was substantive, one was an observation I'm deliberately leaving as is.

**Acted on: the small-batch hole in the failure signal.**

> the failure signal is keyed to the circuit breaker tripping, which requires 3 consecutive transport failures. For a residual batch smaller than that -- most notably `_heal_precomputed_gaps`, which typically fetches just 1-2 missing episodes -- a fully dead provider can never trip the breaker within that call

Correct, and worth more than a spec footnote: it meant the heal path's protection was weakest exactly where that path is most often exercised. `ProviderHealth.failed` now also returns True when a provider **raised on every attempt and never once answered**, which is a verdict the batch's own evidence supports regardless of the threshold.

The `never answered` qualifier is what stops this collapsing into the `transport_failures > 0` rule I rejected in the original design. A provider that answered even once is reachable, so a later blip does not condemn it; a provider the cascade never reached has neither failures nor responses and stays unflagged. Both of those are pinned by tests that already passed before the widening, so they are guards rather than decoration.

New coverage: one- and two-job batches against a dead client, the two over-flag guards above, and the heal path end to end through the **real** scheduler (not a fabricated `SchedulerRun`) with a one-episode gap. The three defect-catching tests were verified to fail against the previous breaker-only rule. Spec doc updated with the reasoning.

**Not acted on: the duplicate log lines.** `run_jobs` logs the generic "never answered" line with failure counts, and the callers log a season-scoped one. They carry different information (counts vs. which show/season is affected) and the second review called it "harmless and arguably useful", so I left both.

Also merged `main` in to clear the conflict with #635 (both sides appended a test class at the same point in `test_build_subtitle_cache.py`; kept both).

`uv run pytest tests/unit`: **3212 passed, 21 skipped**.

PR #649, opened 2026-09-10
feat(subtitle-cache): nightly server harvest with a publish shrink guard

Jsakkos:
## What this does

Phase 3 of the subtitle-cache work: the pieces needed to run the nightly cache harvest on a server instead of by hand on a laptop. This PR adds the repo-side artifacts and the operator runbook. It does not touch any server; that is a manual follow-up the runbook covers.

Phase 1 (the harvester repair) shipped in #631, #635 and #636. The rolling `subtitle-cache-latest` release was separately rebuilt and republished from the laptop corpus on 2026-09-08, going from 467 shows / 36,742 episodes to 479 shows / 37,799 episodes as a strict superset.

## Why

The cache had gone stale for two months because publishing it was a manual chore. Automating it introduces a new risk: an unattended job that publishes a bad artifact is worse than one that publishes nothing, because every Engram install downloads whatever is on that release. Most of this PR is about making the automation refuse to do the wrong thing quietly.

## What is in it

| File | Purpose |
|---|---|
| `backend/scripts/publish_guard.py` | Refuses to publish a cache materially smaller than the live one. 405 lines, 54 unit tests. |
| `deploy/subtitle-cache/harvest.sh` | The nightly wrapper: harvest, pack, guard, publish. |
| `deploy/subtitle-cache/engram-subtitle-cache.{service,timer}` | systemd user units, daily at 02:00 UTC. |
| `deploy/subtitle-cache/engram-subtitle-cache.env.example` | Template for the 0600 credentials file the operator writes. |
| `docs/development/subtitle-cache-server.md` | Install, cutover, monitoring and recovery runbook. |

## Things a reviewer should know

**The wrapper never publishes the build script's own tarball.** `build_subtitle_cache.py` exits 2 when its quota guard halts it partway, and its tarball then holds only the shows completed before the halt. The wrapper instead packs separately from the whole on-disk corpus, so the artifact is complete even on a night the harvest stopped early. This is the single property most likely to be "simplified" away later, so it is documented at the top of the script.

**Exit codes are a contract, not a status.** `harvest.sh` exits 0 (published, complete), 10 (published, harvest halted on quota, the normal nightly outcome), 20 (guard blocked), 21 (guard could not decide), 1 (everything else), 143 (SIGTERM). The service unit carries `SuccessExitStatus=10` so a healthy quota-halted night does not show as a failed unit and train the operator to ignore real failures.

**The guard fails closed.** Several review rounds went into paths where it exited 0 while printing something healthy-looking: a zeroed published baseline disabling every floor, a `gh` outage being indistinguishable from "no baseline exists", a bare `not found` marker matching `HTTP 404: Not Found` auth failures, and the guard reading its baseline from a different repo than the wrapper writes to. Unrecognised failures now resolve to "could not decide" rather than "allow".

**The service unit deliberately has no `[Install]` section.** Enabling it alongside the timer would wire the harvest to `default.target` and fire it on every user-manager start, spending the metered quota more than once a day. The runbook enables only the timer.

**`ProtectHome` is deliberately absent** from the unit and both the unit and the runbook say why: the job reads and writes the checkout, `~/.engram/cache`, `~/.engram/harvest`, and `~/.config/gh`, all under `$HOME`.

## Testing

- 54 unit tests for the guard, all passing. No test makes a network call.
- The guard was run against the real published manifest: `publish-guard: growth: 479 -> 479 shows, 37799 -> 37799 episodes`, exit 0.
- `harvest.sh`: `bash -n` clean, shellcheck clean via `uvx --from shellcheck-py shellcheck`, and the full exit-code matrix (0, 10, 20, 21, 1, 143) driven with stub `uv`, `gh` and `flock` executables. The real harvest was never executed; it spends a metered quota and takes hours.
- Docs build clean with no new warnings.

## Not covered here

The units have not been run against real systemd, and the `TERM` trap's mid-publish behaviour is verified by inspection only. Both get exercised the first time the timer runs on the server.

The server-side install, the corpus and coverage-database cutover, and retiring the laptop harvester are manual steps in the runbook. They need a sudo password (for `loginctl enable-linger`) and the operator's own credentials file.


Jsakkos:
@claude please review this PR

PR #656, opened 2026-09-14
fix(matching): read damaged reference subtitles and refuse seasons with too few usable ones

Jsakkos:
## Summary

A user reported that every track on a Dexter's Laboratory disc matched S01E01, with the same thing happening on a Looney Tunes set. The diagnostics bundle showed 37 reference subtitles for the season, but 36 of them parsed to 0 characters. The one reference with text won every TF-IDF chunk vote, because a lone scoring candidate counts as a clear win, so every track landed on it at full confidence.

The validator only sniffed for `-->`, so these broken files counted as valid, were cached, and were reused on every retry.

This PR fixes the parsing, stops matching when too few references are usable, and lets three-segment cartoon tracks be confirmed as combined.

## Root cause

`SubtitleReader` split on `"\n\n"` and expected the timestamp on the second line of each block. Real files break that in several ways, all found by read-only scans of 37,867 cached SRTs:

- **Doubled line endings.** `\r\r\n` endings, which manual subtitle import on Windows was also creating (`write_text` without `newline=""`).
- **Missing cue index lines.**
- **Short or long timing fields.** Fields with one or three digits (`00:00:0,616`, `00:100:02,227`, `446:12:46,016`), which a two-digit pattern silently dropped.
- **UTF-16 decoding mismatch.** UTF-16 files with a BOM and a stray odd trailing byte: the validator decoded them, but the matcher fell back to latin-1 and read nothing.

## Changes

**Parsing (`srt_utils.py`)**
- One block-based cue parser (`iter_srt_cues`) shared by the matcher and the validator. It measures the gap between timing and text per file, using the most common gap with ties going to the wider one, so mixed layouts parse.
- A shared `decode_utf16_bom`.
- A file is valid when it has at least one cue whose text is not a watermark (`has_dialogue_cues`). Download-link stubs now re-download. Sound-effect-only subtitles are kept, since rejecting them would re-download them forever.

**Matching**
- `TfidfMatcher.prepare` leaves empty references out of the corpus.
- `MIN_USABLE_REFERENCES = 2`: a season with fewer usable references is refused before ffprobe or the ASR model load.
  - The floor also applies to the shipped precomputed cache, since its builders drop empty texts and a season built from damaged subtitles can ship as a single row.
  - It is checked after augmentation.
- Tracks the runtime pre-filter hints are conjoined are scanned at 19 points instead of 10. 19 is the next scan-lattice level, so earlier transcripts are reused. A three-segment track needs more than `3 * 3 + 1` points before `decompose_vote_runs` will confirm it.

**Review routing**
- New error code `references_unreadable`: the title goes to REVIEW unless a DiscDB mapping identifies it.
  - The message says whether the season had too few references or unreadable ones.
  - The code is in `_NON_REMATCHABLE_REVIEW_ERRORS`, so escalation does not re-match it or wipe `match_details`.
- Multi-season matching keeps the refusal only when every season refused.
- `_detect_wrong_show` does not blame a show whose candidates were all refused.
- A re-match that returns no details clears a stale refusal.

**UI**
- The review Inspector shows the reviewer-facing message for `references_unreadable` and `multi_episode_detected`. It uses an allowlist, so raw exception text never shows up there.

## Verification

- `ruff check` and `ruff format --check`: clean.
- Full backend unit tier with `-p no:logging`: 3284 passed, 21 skipped, 0 failed.
  - The 17 errors are the known caplog fixture errors from that flag.
  - The affected files re-run with logging enabled: 408 passed.
- After rebasing onto main, targeted backend files plus `tests/integration/test_subtitle_workflow.py`: all pass.
- `Inspector.test.tsx` (vitest): 26 passed.
- Read-only scan of the real subtitle cache, comparing validator verdict with whether the matcher reads any text:

  | Verdict | Files |
  |---|---|
  | Valid and usable | 37,851 |
  | Valid but unusable | 9 (Primal's sound-effect subtitles, by design) |
  | Invalid but usable | 0 |
  | Invalid and unusable | 7 |

- Across 2,346 show/season groups, the floor fires for 14 seasons. Each of those has exactly one cached reference.

## Notes

- **Rebuild the published subtitle cache once this ships.** The nightly server build still runs the old parser, so seasons built from damaged subtitles may be missing rows.
- **Relation to #614.** Combined `S01E01-E03` filenames are a follow-up (PR 2 in the spec) built on the episode-code design from #614. This PR only makes combined tracks detectable and sends them to review.
- **Follow-ups filed separately:**
  - Move the remaining legacy SRT parsers onto `iter_srt_cues`.
  - Check gating of the `/api/test` router, which is the only path to `MultiSegmentMatcher`.
  - Key the matcher's text caches on file mtime.
- **Squash message.** One commit on the branch carries a stray "Claude Sonnet 5" co-author trailer. Please use "Claude Opus 5" in the squash message.
- **Design docs.** Spec: `docs/superpowers/specs/2026-09-13-empty-srt-references-design.md`. Plan: `docs/superpowers/plans/2026-09-13-empty-srt-references.md`.


Jsakkos:
@claude please review this PR

PR #659, opened 2026-09-16
fix(matcher): parse SRTs in the investigation script via the shared parser

Jsakkos:
## What changed

Two remaining call sites still hand-rolled SRT parsing (split content on `"\n\n"`, require the timing line at `lines[1]`, call `SubtitleReader.parse_timestamp`). Both now read cues through the shared damage-tolerant `iter_srt_cues` from `app/matcher/srt_utils.py` that landed in #656:

- `backend/app/matcher/scripts/generate_investigation_data.py` - the 21-line block loop collapses to 7 lines computing the same two values (max end time, space-joined per-cue text).
- `backend/tests/real_data/test_chunk_vote_real_cache.py` - the `last_end` scan becomes a single `max(..., default=0.0)` generator. The `try/except` and its comment about malformed timestamps go away because `iter_srt_cues` already drops an unparseable timing line internally.

## Why

The hand-rolled form silently misreads files with doubled or tripled line endings (`\r\r\n`), a leading BOM, missing cue index lines, or 1-digit minute/second fields - exactly the damaged-subtitle class #656 was about. An A/B of the old and new logic:

| input | old | new |
|---|---|---|
| well-formed | `(['Hello there friend', 'Second cue'], 9.0)` | identical |
| same file, `\r\r\n` endings | one merged garbage block, duration `4.0` | `(['Hello there friend', 'Second cue'], 9.0)` |

## Reviewer notes

- **Text-less cues are filtered explicitly** (`if cue.lines`). The old `len(lines) < 3` guard did double duty: it required a timing line *and* at least one text line, so a text-less cue never contributed to the duration. `iter_srt_cues` yields those cues with empty `.lines`, so porting without the filter would have quietly changed the computed duration. The filter also matches what `extract_subtitle_chunk` does, keeping the duration scan and the chunk extraction on the same definition of "a cue that counts". This is why `SubtitleReader.get_duration` is not a drop-in here: it deliberately does not filter on `.lines`.
- **No unit test added.** Neither site is a non-script production code path (one is a CLI script, one is a test helper), and `iter_srt_cues` is already covered by `TestIterSrtCues` in `tests/unit/test_srt_utils.py`.
- **`SubtitleReader.parse_timestamp` is now test-only in both copies, and both are left in place** (kept the change minimal - happy to remove them in a follow-up):
  - `episode_identification.py:2301` - only caller left is `tests/unit/test_episode_identification.py:329`. Its sole production caller was the script changed here.
  - `srt_utils.py:249` - zero callers, including tests; it was already dead before this change. The unit test imports `SubtitleReader` from `episode_identification`, not from `srt_utils`.

## Testing

- `uv run ruff check .` clean; both files already ruff-formatted.
- `uv run pytest tests/real_data/test_chunk_vote_real_cache.py tests/unit/test_episode_identification.py tests/unit/test_srt_utils.py` - 135 passed, 2 skipped (the real-data tests auto-skip without the real cache present; nothing wrote to `~/.engram`).


Jsakkos:
@claude please review this PR

PR #660, opened 2026-09-16
fix(security): gate the /api/test/* matcher harness

Jsakkos:
## What changed

The three developer-only matcher endpoints under `/api/test/*` were mounted unconditionally in `main.py` with no DEBUG check, no origin guard and no path validation. The router now carries `Depends(require_debug)` plus `Depends(require_localhost_or_lan)`.

`require_debug` moves from `routes.py` to `guards.py` so `test_routes.py` can take the gate without importing the whole route module, which is the same import-cycle reason the origin gates already live there. `routes.py` imports the name, so its object identity is unchanged and the existing `app.dependency_overrides[require_debug]` pattern in `tests/integration/test_phase2_e2e.py` keeps working (verified: `routes.require_debug is guards.require_debug`).

## Why

All three endpoints take a client-supplied path that reaches ffmpeg and Whisper:

| Endpoint | Input | Reaches |
|---|---|---|
| `POST /api/test/subtitles` | `show_name`, `season` | Outbound scraping (Addic7ed, TVsubtitles, TMDB); writes/deletes SRTs under the config-derived cache dir |
| `POST /api/test/transcribe` | `video_path`, `start_time`, `duration` | `subprocess.run([ffmpeg, "-i", video_path, ...])` in `srt_utils.py:340`, then Whisper |
| `POST /api/test/match` | `video_path`, `show_name`, `season` | `Path(video_path).is_dir()` / `.glob("*.mkv")`, then ASR + matcher over everything found |

Ungated, any peer that could reach the Engram port could have the server read and transcribe an arbitrary file and read the text back, probe the filesystem via the returned `video_duration`, or burn CPU/GPU on demand. `/match` accepts a directory, so one request can enumerate and transcribe a whole tree.

This is **not** command injection: the ffmpeg call is a list argv with no `shell=True`. CORS was configured but is a browser policy, not a server-side gate, so `curl` was unaffected.

## Why both gates

`DEBUG=true` is the documented way to run the dev and E2E backend, so the DEBUG gate alone would leave the surface open on a LAN-exposed dev box. The origin gate alone would leave it live in release builds. Together this is strictly tighter than `/api/simulate/insert-disc-from-staging`, which takes an arbitrary path behind `require_debug` only.

## No callers

`grep` over `frontend/src`, `frontend/e2e`, the backend test suite, scripts and docs found zero references to `/api/test/*` outside the router definition and one passing mention in a design spec. This changes reachability, not an API contract.

## Deliberately not added

No `is_within_configured_roots` check on `video_path`. The harness exists to point the matcher at sample media outside the library roots, so a root check would break its only intended use. Reachability is the control here, and a comment in the file says so.

## Testing

New `backend/tests/unit/test_test_routes_gate.py`, 12 tests: each endpoint 403s with DEBUG off; the gate runs before the handler (`transcribe_chunk.assert_not_called()`); each endpoint 200s from loopback with DEBUG on; each 403s for a LAN peer with `allow_lan_access=False`; the LAN peer is admitted when the user opted in; and the two gates are independent, so LAN opt-in does not unlock a non-debug build.

`httpx.ASGITransport` defaults the peer address to loopback, so the 403 branch is only reachable by passing an explicit `client=("192.168.1.50", 51234)` to the transport.

I also confirmed the tests are sensitive to the gate rather than passing vacuously: with both dependencies overridden off, the request reaches `testing_service.transcribe_chunk` and raises from inside the handler, which is the pre-change behaviour.

- `uv run ruff check .` clean, `ruff format` clean
- `uv run pytest tests/unit/test_test_routes_gate.py tests/unit/test_guards_logging.py tests/unit/test_api_routes.py` 82 passed
- `test_media_endpoints.py test_validation.py test_contributions_localhost.py test_staging_import.py` 154 + 76 passed, covering everything touching the relocated guard

## Reviewer note

Out of scope here, but `/api/simulate/insert-disc-from-staging` and `/api/simulate/trigger-real-scan` also take client-supplied paths behind `require_debug` alone. Same class of surface, weaker gate than `/api/test/*` now has. Worth a follow-up.


Jsakkos:
@claude please review this PR

